PR #39, opened 2014-07-29
pass Array Buffer to constructor

calvinmetcalf:


feross:
We can add support for this feature now that node.js#next has support! https://github.com/nodejs/io.js/pull/2002

feross:
ArrayBuffer can be passed into the constructor as of buffer 3.4.0.

calvinmetcalf:
vindicated! thanks @feross

feross:
@calvinmetcalf Yeah :) I agreed it's a great idea but want to keep this module exactly in sync with node. I argued to support this - happy it finally landed!!

PR #42, opened 2014-09-11
Added support for constructing from ArrayBuffer and added test that verifies support

robskillington:


feross:
We can add support for this feature now that node.js#next has support! https://github.com/nodejs/io.js/pull/2002

feross:
ArrayBuffer can be passed into the constructor as of buffer 3.4.0.

PR #55, opened 2015-02-13
remove redundant code unit size check

jessetane:
e6fddaa9238c815daaf5e61d541a45bc5ef63105 handles this now, so I don't think there is any reason to keep this (from dff29085b184e9c137514b7cc2da422164e89765) hanging around.

### DIFF
--- a/index.js
+++ b/index.js
@@ -371,7 +371,7 @@ function base64Write (buf, string, offset, length) {
 }
 
 function utf16leWrite (buf, string, offset, length) {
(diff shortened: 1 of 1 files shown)


dcousens:
ACK, but is there tests to prove any of this?

jessetane:
@dcousens cbee49cb7e62f19ea44a66c08a9bc9661bc702a5 was the original test, which still passes

feross:
Released as 3.0.3.

PR #57, opened 2015-03-05
index: don't use reserved keyword "byte"

### LINKED ISSUE #56: Bad name of variable
In this place:

347: var byte = parseInt(string.substr(i * 2, 2), 16)
348: if (isNaN(byte)) throw new Error('Invalid hex string')
349: buf[offset + i] = byte

https://github.com/feross/buffer/blob/3a22a49dad272004455e90c1df3b0ffb5a07f47f/index.js#L347-L349

used variable with name "byte", but such names are only allowed in ecmascript 5. I'm use RhinoScriptEngine in java and Rhino not complete support ecmascript 5 now. 
Please change the code for this example:

347: var byte2 = parseInt(string.substr(i * 2, 2), 16)
348: if (isNaN(byte2)) throw new Error('Invalid hex string')
349: buf[offset + i] = byte2

Thank you.

dcousens:
Fixes #56

### DIFF
--- a/index.js
+++ b/index.js
@@ -344,9 +344,9 @@ function hexWrite (buf, string, offset, length) {
     length = strLen / 2
   }
   for (var i = 0; i < length; i++) {
-    var byte = parseInt(string.substr(i * 2, 2), 16)
-    if (isNaN(byte)) throw new Error('Invalid hex string')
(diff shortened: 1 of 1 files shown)


feross:
thanks!

Issue #58, opened 2015-03-11
Pull in iojs changes / perf improvements


feross:
History: https://github.com/iojs/io.js/commits/v1.x/lib/buffer.js

feross:
I just merged all the latest buffer changes from io.js. Almost all the changes were related to performance.

Released as 3.2.0.

feross:
Pretty nice performance improvement to the `Buffer` constructor. :+1: 

Before:
BrowserBuffer#new(16) x 863,493 ops/sec ±1.61% (93 runs sampled)

After:
BrowserBuffer#new(16) x 1,249,470 ops/sec ±0.96% (89 runs sampled)

dcousens:
Cheers for writing this up :)

PR #59, opened 2015-04-22
Cleanup

dcousens:
Some follow ups to https://github.com/feross/buffer/issues/58

### DIFF
--- a/index.js
+++ b/index.js
@@ -1095,11 +1095,11 @@ Buffer.prototype.writeDoubleBE = function writeDoubleBE (value, offset, noAssert
 }
 
 // copy(targetBuffer, targetStart=0, sourceStart=0, sourceEnd=buffer.length)
-Buffer.prototype.copy = function copy (target, target_start, start, end) {
+Buffer.prototype.copy = function copy (target, targetStart, start, end) {
   if (!start) start = 0
   if (!end && end !== 0) end = this.length
-  if (target_start >= target.length) target_start = target.length
-  if (!target_start) target_start = 0
+  if (targetStart >= target.length) targetStart = target.length
+  if (!targetStart) targetStart = 0
   if (end > 0 && end < start) end = start
 
   // Copy 0 bytes; we're done
@@ -1107,7 +1107,7 @@ Buffer.prototype.copy = function copy (target, target_start, start, end) {
   if (target.length === 0 || this.length === 0) return 0
 
   // Fatal error conditions
-  if (target_start < 0) {
+  if (targetStart < 0) {
     throw new RangeError('targetStart out of bounds')
   }
   if (start < 0 || start >= this.length) throw new RangeError('sourceStart out of bounds')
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -105,7 +105,7 @@ function fromString (that, string, encoding) {
   var length = byteLength(string, encoding) | 0
   that = allocate(that, length)
 
-  that.write(string, encoding) | 0
+  that.write(string, encoding)
```
dcousens: Unused result?  Why `| 0`?

index.js
```diff
@@ -114,7 +114,7 @@ function fromObject (that, object) {
 
   if (isArray(object)) return fromArray(that, object)
 
-  if (object == null) {
+  if (object === null || object === undefined) {
```
dcousens: This is [not a performance issue](https://jsperf.com/check-for-null-undefined).
When I made this change, I was under the impression `==` was not allowed in standard.  Apparently it is,  so I'm happy to switch this back.

I personally still prefer the above as a more verbose equivalent, or I prefer the `if (!object)` approach.  Up to you @feross.

index.js
```diff
@@ -855,15 +864,15 @@ Buffer.prototype.writeUIntLE = function writeUIntLE (value, offset, byteLength,
...
-  byteLength = byteLength >>> 0
+  offset = offset | 0
+  byteLength = byteLength | 0
   if (!noAssert) checkInt(this, value, offset, byteLength, Math.pow(2, 8 * byteLength), 0)
 
   var i = byteLength - 1
   var mul = 1
   this[offset + i] = value & 0xFF
   while (--i >= 0 && (mul *= 0x100)) {
-    this[offset + i] = (value / mul) >>> 0 & 0xFF
```
dcousens: `>>> 0` was superfluous here.

index.js
```diff
@@ -230,11 +230,20 @@ Buffer.compare = function compare (a, b) {
 
   var x = a.length
   var y = b.length
-  for (var i = 0, len = Math.min(x, y); i < len && a[i] === b[i]; i++) {}
```
dcousens: No performance issue, but the intention is less clear and IMHO misleading.

index.js
feross: I think `== null` is fine.

index.js
```diff
@@ -206,7 +206,7 @@ function checked (length) {
     throw new RangeError('Attempt to allocate Buffer larger than maximum ' +
                          'size: 0x' + kMaxLength.toString(16) + ' bytes')
   }
-  return length >>> 0
+  return length | 0
```
feross: Why? I'd rather keep this the same as the iojs codebase so it's easier to stay in sync.

index.js
```diff
@@ -230,11 +230,20 @@ Buffer.compare = function compare (a, b) {
 
   var x = a.length
   var y = b.length
-  for (var i = 0, len = Math.min(x, y); i < len && a[i] === b[i]; i++) {}
+
+  var i = 0
+  var len = Math.min(x, y)
+  while (i < len) {
+    if (a[i] !== b[i]) break
```
feross: Why not `while (i < len && a[i] === b[i]) {`?

feross: I guess it's a bit clearer the way you wrote it.

index.js
feross: Good catch.

index.js
dcousens: Consistency?  I'm happy to try and merger this back up stream.

feross:
Tests failed. Rerunning.

index.js
feross: If you're interested, go for it.

feross:
Tests passed. I rebased to remove your `== null` changes and I'll publish a new patch version as soon as my local tests pass.

dcousens:
Sounds good :)

Cheers @feross

feross:
Published as 3.2.1.

dcousens:
By the way, this PR's branch was in this repository, feel free to collaborate/rebase the PR's branch in future rather than rebasing externally.  

Something satisfying about that purple "success" after a PR merge haha.  Also more visible from a notifications POV.

feross:
Will do next time! Thanks :-)

PR #64, opened 2015-08-04
Fix base 64 character exclusion regex 

daverayment:
Using [A-z] in the regex brings in non-alphanumeric characters between the uppercase and lowercase alphabets, meaning the following chars were not correctly stripped from base 64 strings: left and right square bracket, backslash, caret, underline, and backtick. Explicitly separating the uppercase and lowercase letter ranges fixes this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1290,7 +1290,7 @@ Buffer._augment = function _augment (arr) {
   return arr
 }
 
-var INVALID_BASE64_RE = /[^+\/0-9A-z\-]/g
+var INVALID_BASE64_RE = /[^+\/0-9A-Za-z\-]/g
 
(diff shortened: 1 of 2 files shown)


feross:
Thanks for the bug fix and test to verify it. This was failing another test. Specifically, it broke support for url-safe base64. You can run the tests locally with `npm test`.

`_` is an allowed character that is used in "url safe" base 64, which node supports, so I corrected the list of valid characters in https://github.com/feross/buffer/commit/d3619243c0e3ac800cbeef1f41bdb63092b3c83b,

Released as 3.3.2.

daverayment:
Thanks, @feross, and apologies for missing that regression issue.

PR #65, opened 2015-08-06
descending copy from end

kawanet:
Node.js's native Buffer module uses descending copying order in case that target buffer overlaps source buffer. This patch simulates the same behavior.

```
  var b = new Buffer('abcdefghij')
  b.copy(b, 3, 0, 7) // => 'abcabcdefg' but not 'abcabcabca'
```


dcousens:
LGTM.

feross:
Released as 3.4.1.

kawanet:
😄

Issue #66, opened 2015-08-10
Different result than with node


yehudahkay:
Hi,

I tried using specifically the function utf8slice on a byte array but I get a slightly different result than calling new Buffer(arr).toString() from node. I tried finding the utf8slice function in the node source code but didn't have much luck. 

Here is the function I used from your source code:
```
var res = ''
    var tmp = ''
   var end = arr.length

    for (var i = 0; i < end; i++) {
        if (arr[i] <= 0x7F) {
            res += decodeUtf8Char(tmp) + String.fromCharCode(arr[i])
            tmp = ''
        } else {
            tmp += '%' + arr[i].toString(16)
        }
    }

    var str = res + decodeUtf8Char(tmp)

    function decodeUtf8Char(str) {
        try {
            return decodeURIComponent(str)
        } catch (err) {
            return String.fromCharCode(0xFFFD) // UTF 8 invalid char
        }
    }
```
Here is the array:
```
[...]

PR #67, opened 2015-08-11
Insert replacement characters like node does when converting UTF8 bytes to String

jessetane:
Possible fix for #66


dcousens:
Will review today

jessetane:
Cool, thanks. I just installed node@3 and I see now it is replacing things in the surrogate pair range - pushing a test and fix for that now..

feross:
LGTM. Waiting for @dcousens. Feel free to merge when it looks good. Then I'll do a release.

dcousens:
LGTM.  Nice implementation,  I might break down down some of the indentation through the use of `continue`/`break`.

feross:
Published as 3.4.2.

Issue #69, opened 2015-08-15
"call stack size exceeded" for large binary strings in 3.4.2


nolanlawson:
In PouchDB we use large binary strings when we store binary data in `localstorage-down` and `level-js`, and our tests started failing recently due to buffer 3.4.1 and 3.4.2 (https://github.com/pouchdb/pouchdb/issues/4164, see discussion over there). In short, 3.4.1 had a perf regression which appears fixed in 3.4.2, but 3.4.2 now has a "call stack size exceeded" error when I test in Chrome 44 (on an admittedly slow laptop, so it may be hard to reproduce on a faster one).

Glancing at the code, I think a reasonable fix would be to break up the array into decent-size chunks and then concat the string together, but I admit I'm not super familiar with the codebase.

jessetane:
Ah, yes the `apply` argument count limit... Chrome seems to bail for > 65535. I had originally written this calling `String.fromCharCode` for each code point but that was rather slow, so @nolanlawson I think your idea to break up the array into chunks at the end seems maybe best.

dcousens:
Resolved in #70

adametry:
Unfortunately, there's not a fixed max arguments (i.e., 0x10000 in Chrome) as the size/complexity of the call stack also influences this threshold. I receive the "call stack size exceeded" while converting moderately large buffers to strings before (@3.4.2) and after the c5a9012 commit. I've wrestled with this issue before and found it best to either use a much smaller chunk size (e.g., 0xfff) or avoid chunked applies entirely. Though, I have no doubt that there are more brilliant solutions. 

To better understand the what is happening, here are two helpful SO explanations:
• [“RangeError: Maximum call stack size exceeded” Why?](http://stackoverflow.com/a/22123980)
• [Chrome: How to solve “Maximum call stack size exceeded” errors on Math.max.apply( Math, array )](http://stackoverflow.com/a/20940386)

nolanlawson:
@adametry Can you open a new issue then? I think using a smaller chunk size is probably a fine solution. I'd suggest a try/catch, but that would lead to deoptimizations, so probably worse.

feross:
We should try making the max argument length smaller - say cut it in half. If that doesn't work, we should probably stop using the `apply` approach.

dcousens:
@feross agreed,  want me to make the change?

jessetane:
[...]

PR #70, opened 2015-08-15
Fix call stack exceeded in #69

nolanlawson:
Based on [a JSPerf](http://jsperf.com/string-fromcharcode-apply-vs-string-fromcharcode-using-), calling `String.fromCharCode.apply(String, array)` is indeed faster than concatenating a string in a for-loop. But of course, even slow performance is preferable to a runtime error. :)

I haven't tested to find the optimal cutoff, but 0x10000 seems like a decent size for a "large" array. For comparison, we have a test in PouchDB that puts some large PNG data into localStorage as a binary string, and in buffer v3.4.0 it was taking ~45s to run, whereas using this fix we are down to 2.7s (and it is passing, unlike v3.4.1).

### DIFF
--- a/index.js
+++ b/index.js
@@ -622,6 +622,28 @@ function base64Slice (buf, start, end) {
   }
 }
 
+function decodeLargeCodePointsArray (array) {
+  var res = ''
+  var i
+  var end = array.length
+
+  for (i = 0; i < end; i++) {
+    res += String.fromCharCode(array[i])
+  }
+
+  return res
+}
+
+function decodeCodePointsArray (array) {
+  if (array.length < 0x10000) {
+    return String.fromCharCode.apply(String, array)
+  }
+  // Decode using string concatenation to avoid "call stack size exceeded".
+  // Based on http://stackoverflow.com/a/22747272/680742, the browser with
+  // the lowest limit is Chrome, with 0x10000 args.
+  return decodeLargeCodePointsArray(array)
+}
+
 function utf8Slice (buf, start, end) {
   end = Math.min(buf.length, end)
   var firstByte
@@ -700,7 +722,7 @@ function utf8Slice (buf, start, end) {
     res.push(codePoint)
   }
 
-  return String.fromCharCode.apply(String, res)
+  return decodeCodePointsArray(res)
 }
 
 function asciiSlice (buf, start, end) {


jessetane:
Did you try using `Array.slice` to limit the chunk size so that we're at least always using the "fast" way? There should be a benchmark you can use to check `perf/readUtf8.js`.

Also maybe useful: http://stackoverflow.com/questions/22747068/is-there-a-max-number-of-arguments-javascript-functions-can-accept

nolanlawson:
I considered `Array.slice`, but my hunch is that calling `slice()` a bunch of times (and thus copying arrays) would be less performant than just using a for-loop. Need a JSPerf to back that up, though. :)

jessetane:
yeah, and maybe that's micro optimizing, generally tho lgtm

nolanlawson:
Chrome 33.0.1750.154 m: The last successful test was 65535 arguments

Ooh, I am good; I guessed 0x10000. :D

nolanlawson:
Style test is failing; my bad. I will fix and then add a comment with the link you posted; Chrome has the lowest limit and it is indeed 0x10000.

index.js
```diff
@@ -622,6 +622,28 @@ function base64Slice (buf, start, end) {
   }
 }
 
+function decodeLargeCodePointsArray (array) {
+  var res = ''
+  var i
```
dcousens: Nit: why bother defining `i` here?

dcousens:
Thanks @nolanlawson ,  nice PR!

index.js
nolanlawson: I was following the style of the rest of the code; I thought y'all were var-hoisters. :)

nolanlawson:
Well I'll be damned; looks like `slice()` in chunks is indeed faster: http://jsperf.com/string-fromcharcode-apply-vs-string-fromcharcode-using-/3

I'll write a follow-up PR. :)

dcousens:
@nolanlawson sounds good! :),  I removed the function in https://github.com/feross/buffer/commit/3d622d3c0bf330e332d6656db7aa3880adb52566,  so make your optimizations to `binarySlice` instead :+1:

nolanlawson:
ah okay, will do

PR #72, opened 2015-08-15
Improve perf of fromCharCode(), fixes #71

### LINKED ISSUE #71: Improve perf of fromCharCode()
Follow-up to #69 and #70; looks like chunking up the array is actually faster: http://jsperf.com/string-fromcharcode-apply-vs-string-fromcharcode-using-/3

nolanlawson:
Cool, that test of ours I mentioned in #70 (the "big png" test) went from 2.7s to 1.4s. Thanks @jessetane for pointing this out!

### DIFF
--- a/index.js
+++ b/index.js
@@ -622,14 +622,14 @@ function base64Slice (buf, start, end) {
   }
 }
 
-function decodeCodePointsArray (array) {
-  // Based on http://stackoverflow.com/a/22747272/680742, the browser with
-  // the lowest argument limit is Chrome, with 0x10000 args.
-  if (array.length < 0x10000) return String.fromCharCode.apply(String, array)
+function decodeCodePointsArray (buf) {
+  var len = buf.length
 
-  // If above that limit, decode using string concatenation
-  // to avoid "call stack size exceeded".
-  return binarySlice(array, 0, array.length)
+  if (len <= 0x10000) {
+    return String.fromCharCode.apply(String, buf) // avoid extra slice()
+  }
+
+  return binarySlice(buf, 0, len)
 }
 
 function utf8Slice (buf, start, end) {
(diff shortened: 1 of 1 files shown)


dcousens:
@nolanlawson any reason you can't apply this optimization to `binarySlice`?

nolanlawson:
OK, done

index.js
```diff
@@ -622,14 +622,14 @@ function base64Slice (buf, start, end) {
...
 }
 
-function decodeCodePointsArray (array) {
-  // Based on http://stackoverflow.com/a/22747272/680742, the browser with
-  // the lowest argument limit is Chrome, with 0x10000 args.
-  if (array.length < 0x10000) return String.fromCharCode.apply(String, array)
+function decodeCodePointsArray (buf) {
+  var len = buf.length
+
+  if (len <= 0x10000) {
```
dcousens: I might pull this constant out and put the explanation comment on it,  that way its not magically here.

dcousens:
Thanks @nolanlawson

index.js
nolanlawson: thanks, that's a good idea!

nolanlawson:
No prob! Any chance of a 3.4.3 release soon? Would unbreak some of our tests in PouchDB.

index.js
dcousens: Should just merge all this code into `binarySlice`?
I wonder if the manual string concatenation is slower than `String.fromCharCode.apply(String, buf.slice(start, end))`.
@nolanlawson reckon you could find out?

nolanlawson: The issue isn't manual string concatenation; it's doing an unnecessary `Array.slice()` when the buffer.length is < 0x10000. I think it's pretty clear that doing `slice()` is slower than not doing `slice()`, although a JSPerf would tell us exactly how slow.

nolanlawson: I can indeed verify that slice is slower than no-slice: http://jsperf.com/string-fromcharcode-apply-vs-string-fromcharcode-using-/4

index.js
```diff
@@ -723,13 +723,17 @@ function asciiSlice (buf, start, end) {
 }
 
 function binarySlice (buf, start, end) {
-  var ret = ''
-  end = Math.min(buf.length, end)
```
dcousens: Will re-add this,  shouldn't have been removed. 
Probably need to add a test for it.

nolanlawson: ah, my bad. sorry!

PR #73, opened 2015-08-16
utf8Slice: don't replace the replacement character

jessetane:
Noticed another inconsistency with node's Buffer - I hadn't thought about what would happen if there were already replacement chars in the utf8 we were converting!


dcousens:
@jessetane awesome!  I was actually re-writing this,  and this had been a source of confusion for me.

dcousens:
This makes the logic way simpler to follow.

PR #75, opened 2015-08-22
Fix tests

jessetane:
the changes to binary slice appear to be causing some browsers to hang while running the tests, not sure why but putting the old one back seems to fix.

### DIFF
--- a/index.js
+++ b/index.js
@@ -692,7 +692,30 @@ function utf8Slice (buf, start, end) {
     i += bytesPerSequence
   }
 
-  return binarySlice(res, 0, res.length)
+  return decodeCodePointsArray(res)
+}
+
+// Based on http://stackoverflow.com/a/22747272/680742, the browser with
+// the lowest limit is Chrome, with 0x10000 args.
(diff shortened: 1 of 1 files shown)


feross:
Looks like this fixes the issue. I'm going to do a release.

Ideally, we can figure out how to merge these two functions in a way that still works in IE8 for a future version.

feross:
Released as 3.4.3.

Also, added you as a GitHub collaborator @nolanlawson. Good work!

nolanlawson:
Thanks! I'll try to be more careful in the future to avoid regressions. Looks like Sauce tests run for every commit?

feross:
Yep!

dcousens:
Can we get some test fixtures that are reproducible?

feross:
This was already caught by the tests.

The Travis tests run automatically on every push. The badge was red.

You can also run the tests locally by creating a free sauce labs account
(Open Sauce), adding the credentials to your local zuul config file, and
running npm test.

PR #82, opened 2015-11-12
Fix fatal error of surrogte pair

hakatashi:
Current implementation of surrogate pair doesn't take codepoints over U+20000 into consideration. This is invalid since original specification of UTF-16 is valid until U+10FFFF.


dcousens:
@feross LGTM

dcousens:
Thanks @hakatashi

feross:
Published as 3.5.2 – thanks!

PR #83, opened 2015-11-13
avoid copy with single byte per element typed arrays

calvinmetcalf:


feross:
I'm pretty sure that this would break compatibility with node.js because the Buffer constructor always does a copy when you pass in an Array-like object (i.e. something with a `length` property).

I'm surprised that we don't have a test to catch this. I'll add one.

You can see that the backing memory is not shared in node:

```js
var u = new Uint8Array(4)
var b = new Buffer(u)
b[0] = 99
console.log(b) // <Buffer 64 00 00 00>
console.log(u) // Uint8Array { '0': 0, '1': 0, '2': 0, '3': 0 }
```

feross:
Here's the test: https://github.com/feross/buffer/commit/7880f8cc2108056503c6eae28d8531a076df9448

I just confirmed that this merging this PR would fail this test.

feross:
Btw, you can use this package ([typedarray-to-buffer](https://www.npmjs.com/package/typedarray-to-buffer)) to achieve a zero-copy buffer creation.

Going to close this PR now. Thanks for trying :)

calvinmetcalf:
Ok so I just have to pass in the .buffer property if I want to avoid the
copy

feross:
That's right. Or, if you really need to only use a view on the .buffer (i.e. a certain range), you can use `typedarray-to-buffer` which can do that for you.

PR #84, opened 2015-11-30
jspm configuration support

guybedford:
This PR includes the necessary jspm configuration for this package to work both client and server in jspm and supporting the conditional package map feature being released in jspm 0.17.

I completely understand if you'd rather not support this here, but it will simplify user configuration locally having it in one place. There is no maintenance burden either as I would continue to maintain this personally.

Just let me know if you have any questions at all, and thanks for considering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,5 +63,12 @@
       "test/_polyfill.js",
       "perf/*.js"
     ]
+  },
+  "jspm": {
+    "map": {
+      "./index.js": {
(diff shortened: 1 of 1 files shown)


feross:
I don't see a problem with this, if you maintain it.

feross:
Released as 3.5.3.

guybedford:
Excellent! Thank you so much for being open to this.

PR #85, opened 2015-12-05
Use isarray instead of is-array

jscissr:
Because isarray is 10x more popular.
They do exactly the same thing.
`npm i browserify` for example installs them both, with this change only one is installed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,7 +8,7 @@
 
 var base64 = require('base64-js')
 var ieee754 = require('ieee754')
-var isArray = require('is-array')
+var isArray = require('isarray')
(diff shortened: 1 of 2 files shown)


feross:
LGTM. Released as 3.5.4.

dcousens:
Should bump to `1.0.0` if @juliangruber is up for,  ref: https://github.com/juliangruber/isarray/issues/1

dcousens:
`isarray` is now at `1.0.0`

feross:
@dcousens Released as 3.5.5.

PR #88, opened 2015-12-24
Proper strict mode support. Fixes #79

### LINKED ISSUE #79: Cannot set property length of [object Object]
Buffer 3.5.0 breaks in Chrome (`45.0.2454.85 (Official Build) (64-bit)`)

What I'm getting is 
~~~
Uncaught TypeError: Cannot set property length of [object Object] which has only a getter
~~~

It happens when buffer tries to set `.prototype.length` to `undefined`, here: https://github.com/feross/buffer/blob/master/index.js#L372

I discovered this while I was rolling out our app's new version. Later I'm gonna try to put together a small demo.

(I'm using browserify 11.11.0 which uses `buffer@3.5.0`)

nolanlawson:
Adds `'strict mode'` to the beginning of `index.js`,
and makes sure that only strict-compliant operations
are used throughout the code.

This works for me in Chrome 49. Opening a pull request
to test in multiple browsers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,8 @@
  */
 /* eslint-disable no-proto */
 
+'use strict'
+
 var base64 = require('base64-js')
 var ieee754 = require('ieee754')
 var isArray = require('isarray')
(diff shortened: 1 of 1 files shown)


feross:
Awesome, thanks for getting to the bottom of this. Released as 3.6.0.

nolanlawson:
No prob! It was actually kinda fun. :)

PR #91, opened 2016-01-02
Removed Buffer.toArrayBuffer(). Fixes #90

### LINKED ISSUE #90: Remove buffer.toArrayBuffer()
Node removed the buffer.toArrayBuffer() API sometime during node 0.11. Users who rely on it will have code that doesn't work in node.js, which is unfortunate.

Let's remove buffer.toArrayBuffer() so we have parity with node.js.

RaulTsc:


feross:
LGTM

feross:
Released as 4.0.0.

dcousens:
ACK

Issue #40, opened 2014-08-10
Investigate proper `instanceof Buffer` support


feross:
I saw a potential way to get `instanceof Buffer` working in a module by @dcodeIO: https://github.com/dcodeIO/node-BufferView/blob/master/BufferView.js

```js
    // Extend DataView and clear all instance methods to be sure
    BufferView.prototype = Object.create(DataView.prototype);
    (function() { for (var i in BufferView.prototype) delete BufferView.prototype[i]; })();
```

feross:
New possible solution: https://github.com/iojs/io.js/issues/106#issuecomment-76251254

feross:
`instanceof Buffer` works correctly in buffer v3.5.0 now!

This fixes the last remaining incompatibility with the node Buffer! And since Buffer in node v4.0.0 is based on `Uint8Array` as well, we're on exactly the same page now!

There's a 50% performance hit on the 16 byte buffer test (`new Buffer(16)`) in non-V8 engines (much better than my previous experiments to get this working). For the 16KB large buffer test (`new Buffer(1600)`), the hit is only 25%, since the difference is drowned out by the zeroing out operation. I expect this to get better over time since subclassing arrays is supported by ES6 now. :+1:

dcousens:
@feross any reason why `node 4.0.0` couldn't be using the same `Buffer` code as this library? (hell, why not the same library).

feross:
Well, they do a bunch of calls directly into the C++ engine. I think one reason they do that is so they can disable the behavior where new typed array memory must be zeroed-out for security reasons in the browser environment. In node, they can do without that and get perf improvements.

Maybe there are more reasons...

PR #92, opened 2016-01-08
Speed up buffer creation.

mtth:
Since 1d20f50f8b4c62779c0dbfb1fa401168ab52dce0 (#40), instances already gained all the `Buffer`'s
methods via  prototype re-assignment. Skipping the `_augment`
call yields significant speedups:

+ ~50% faster small buffer creation:

  + `new(16)` before x 595,757 ops/sec ±0.92% (91 runs sampled)
  + `new(16)` after x 894,462 ops/sec ±0.69% (87 runs sampled)

+ ~20% faster slice:

  + `slice` before: 2,176,194 ops/sec ±1.59% (89 runs sampled)
  + `slice` after: 2,624,209 ops/sec ±0.42% (99 runs sampled)

Other benchmarks aren't significantly different. (These numbers are for v8.)

Also removed the similarly redundant `_isBuffer` property. In case you'd
like to keep it (even though it's private), let me know.

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,10 +186,11 @@ function fromArrayBuffer (that, array) {
   if (Buffer.TYPED_ARRAY_SUPPORT) {
     // Return an augmented `Uint8Array` instance, for best performance
     array.byteLength
-    that = Buffer._augment(new Uint8Array(array))
+    that = new Uint8Array(array)
+    that.__proto__ = Buffer.prototype
   } else {
     // Fallback: Return an object instance of the Buffer class
-    that = fromTypedArray(that, new Uint8Array(array))
+    that = fromTypedArray(that, array)
   }
   return that
 }
@@ -223,6 +224,7 @@ function fromJsonObject (that, object) {
 
 if (Buffer.TYPED_ARRAY_SUPPORT) {
   Buffer.prototype.__proto__ = Uint8Array.prototype
+  Buffer.prototype._set = Uint8Array.prototype.set // For `copy` below.
   Buffer.__proto__ = Uint8Array
 } else {
   // pre-set for values that may exist in the future
@@ -233,12 +235,11 @@ if (Buffer.TYPED_ARRAY_SUPPORT) {
 function allocate (that, length) {
   if (Buffer.TYPED_ARRAY_SUPPORT) {
     // Return an augmented `Uint8Array` instance, for best performance
-    that = Buffer._augment(new Uint8Array(length))
+    that = new Uint8Array(length)
     that.__proto__ = Buffer.prototype
   } else {
     // Fallback: Return an object instance of the Buffer class
     that.length = length
-    that._isBuffer = true
   }
 
   var fromPool = length !== 0 && length <= Buffer.poolSize >>> 1
@@ -420,6 +421,10 @@ function slowToString (encoding, start, end) {
     }
   }
 }
+
+// Even though this property is private, it shouldn't be removed because it is
+// used by `is-buffer` to detect buffer instances in Safari 5-7.
+Buffer.prototype._isBuffer = true
 
 Buffer.prototype.toString = function toString () {
   var length = this.length | 0
@@ -798,7 +803,8 @@ Buffer.prototype.slice = function slice (start, end) {
 
   var newBuf
   if (Buffer.TYPED_ARRAY_SUPPORT) {
-    newBuf = Buffer._augment(this.subarray(start, end))
+    newBuf = this.subarray(start, end)
+    newBuf.__proto__ = Buffer.prototype
   } else {
     var sliceLen = end - start
     newBuf = new Buffer(sliceLen, undefined)
@@ -1322,6 +1328,9 @@ var BP = Buffer.prototype
 
 /**
  * Augment a Uint8Array *instance* (not the Uint8Array class!) with Buffer methods
+ *
+ * This function is also used externally by `typedarray-to-buffer`.
+ *
  */
 Buffer._augment = function _augment (arr) {
   arr.constructor = Buffer


dcousens:
LGTM, but I think @feross would be the most knowledgeable about all the subtleties for this.

mtth:
Thanks @dcousens !

@feross  - I'm also curious if you would consider adding an option to not rely on prototype mutation (falling back to "augmenting" array instances). Without it, buffers are still unreasonably slow in some (most?) engines. For example here are results for Firefox 43:

+ BrowserBuffer#bracket-notation x 487,929 ops/sec ±1.33% (59 runs sampled) # before
+ BrowserBuffer#bracket-notation x 494,476 ops/sec ±1.19% (66 runs sampled) # after
+  Uint8Array#bracket-notation x 24,571,591 ops/sec ±1.12% (65 runs sampled) # !

I initially thought that clients would just be able to set `window.TYPED_ARRAY_SUPPORT` to false in these cases, but it feels a bit like a hack (since these browsers do support it), and it's slower (especially for slices, which then require a copy).

If this sounds reasonable to you, I'll update this pull request with a tentative implementation.

feross:
We can't get rid of the `_isBuffer` property since that's needed for `is-buffer` (https://github.com/feross/is-buffer) to work in older Safari. `is-buffer` is used by browserify whenever there's a `Buffer.isBuffer` call (very common) to save requiring the whole buffer module in browser bundles.

Removing `_augment` will break typedarray-to-buffer (https://github.com/feross/typedarray-to-buffer) but that can be fixed by moving `_augment` directly into that package.

> I'm also curious whether you would consider adding an option to not rely on prototype mutation (falling back to "augmenting" array instances)

I don't want to expose any options that aren't in the node buffer, because then we'll have code that's browser-specific and could end up doing something different in future versions of node. We should solve the issue directly in this module, but that should be done in another issue/PR, IMO.

mtth:
> We can't get rid of the _isBuffer property since that's needed for is-buffer (https://github.com/feross/is-buffer) to work in older Safari.

Sounds good. I'll add the `_isBuffer` property back.

> Removing _augment will break typedarray-to-buffer (https://github.com/feross/typedarray-to-buffer) but that can be fixed by moving _augment directly into that package.

I'll restore `_augment` as well then. (It might also come in handy if some fallback mechanism is implemented.)

> I don't want to expose any options that aren't in the node buffer, because then we'll have code that's browser-specific and could end up doing something different in future versions of node. We should solve the issue directly in this module, but that should be done in another issue/PR, IMO.

I'm fine doing the detection inside this library (though there's the risk of being suboptimal for future browser implementations). Also, maybe I didn't explain myself correctly, I didn't mean an option at instantiation time but rather something like what `TYPED_ARRAY_SUPPORT` currently does.

index.js
```diff
@@ -1322,6 +1328,9 @@ var BP = Buffer.prototype
 
 /**
  * Augment a Uint8Array *instance* (not the Uint8Array class!) with Buffer methods
+ *
+ * This function is also used externally by `typedarray-to-buffer`.
```
dcousens: :+1:  this is good.

index.js
```diff
@@ -421,6 +422,10 @@ function slowToString (encoding, start, end) {
   }
 }
 
+// Even though this property is private, it shouldn't be removed because it is
+// used by `is-buffer` to detect buffer instances in Safari 5-7.
+Buffer.prototype._isBuffer = true
```
dcousens: This was only being set in the fallback case originally, why is it now *always* set?

mtth: It was already always set (since it's [added inside `Buffer._augment`](https://github.com/feross/buffer/blob/master/index.js#L1328) which was always called). If it wasn't the case then `Buffer.isBuffer` would've failed since it [uses this property](https://github.com/feross/buffer/blob/master/index.js#L269). I just moved it to the prototype to avoid having to add it everywhere.

dcousens: Why is it still added inside `Buffer._augment` then?

dcousens: Nevermind,  great work @mtth

index.js
```diff
@@ -186,10 +186,11 @@ function fromArrayBuffer (that, array) {
   if (Buffer.TYPED_ARRAY_SUPPORT) {
     // Return an augmented `Uint8Array` instance, for best performance
     array.byteLength
-    that = Buffer._augment(new Uint8Array(array))
+    that = new Uint8Array(array)
+    that.__proto__ = Buffer.prototype
   } else {
     // Fallback: Return an object instance of the Buffer class
-    that = fromTypedArray(that, new Uint8Array(array))
+    that = fromTypedArray(that, array)
```
feross: This will break if the user passes an ArrayBuffer into the Buffer object implementation. `fromTypedArray` takes an array-like object.

index.js
```diff
@@ -223,6 +224,7 @@ function fromJsonObject (that, object) {
 
 if (Buffer.TYPED_ARRAY_SUPPORT) {
   Buffer.prototype.__proto__ = Uint8Array.prototype
+  Buffer.prototype._set = Uint8Array.prototype.set // For `copy` below.
```
feross: I know you just copied this but I wonder if we can't just use `Uint8Array.prototype.set.call` instead of saving a reference here.

index.js
```diff
@@ -186,10 +186,11 @@ function fromArrayBuffer (that, array) {
   if (Buffer.TYPED_ARRAY_SUPPORT) {
     // Return an augmented `Uint8Array` instance, for best performance
     array.byteLength
```
feross: I think this is no-op that we can remove. Just noticed it, lol

feross:
I can fix up the issues and merge. I'll add new tests to catch the issues I manually caught.

index.js
```diff
@@ -1322,6 +1328,9 @@ var BP = Buffer.prototype
 
 /**
  * Augment a Uint8Array *instance* (not the Uint8Array class!) with Buffer methods
+ *
+ * This function is also used externally by `typedarray-to-buffer`.
+ *
  */
 Buffer._augment = function _augment (arr) {
   arr.constructor = Buffer
```
feross: I think we still need this line, or `is-buffer` will break.

index.js
feross: Actually, never mind. Heh. I remember why I added this. I'll add a comment.

index.js
feross: Seems to work. I'll make the change.

index.js
feross: Nevermind. This is inherited from `Buffer.prototype`.

feross:
We can actually just remove `Buffer._augment` entirely. I'll update `typedarray-to-buffer` to do just modify `__proto__` like package now does.

index.js
mtth: Thinking about it a bit more, I think we actually need this line. Not for this package, but for `typedarray-to-buffer`. (This doesn't apply if you make the change you mention below about switching to setting the prototype.)

feross: `typedarray-to-buffer` is switched over now, so we don't need this line anymore.

mtth: Great!

feross:
Released as v4.1.0.

mtth:
Thanks @feross @dcousens !

feross:
This is released in browserify v13: https://github.com/substack/node-browserify/blob/master/changelog.markdown

index.js
dcousens: Now you have me curious :)

dcousens:
Awesome work @mtth / @feross  :+1:

Issue #94, opened 2016-01-13
Broken in Chrome 49.0.2620.0 canary (64-bit)


LinusU:
My chrome updated during the night and now my stuff doesn't work anymore, fun times, I get to debug!

The first thing was actually already fixed by 5f5478a34e8ed7b9642407b062e8c5edc6799510, I just had to update to the latest version. @feross Was that commit intentionally fixing a Chrome bug?

The next problem is that `.subarray(start, end)` seems broken when the `__proto__` is tampered with.

![screen shot 2016-01-13 at 16 54 53](https://cloud.githubusercontent.com/assets/189580/12299304/65a3297a-ba16-11e5-98de-3d584309ecc5.png)

The following code fixes the problem but I'm not sure that it's the right approach:

```diff
diff --git a/index.js b/index.js
index 15d7c50..cbb1516 100644
--- a/index.js
+++ b/index.js
@@ -782,7 +782,9 @@ Buffer.prototype.slice = function slice (start, end) {
 
   var newBuf
   if (Buffer.TYPED_ARRAY_SUPPORT) {
+    this.__proto__ = Uint8Array.prototype
     newBuf = this.subarray(start, end)
+    this.__proto__ = Buffer.prototype
     newBuf.__proto__ = Buffer.prototype
   } else {
     var sliceLen = end - start
```

feross:
No, https://github.com/feross/buffer/commit/5f5478a34e8ed7b9642407b062e8c5edc6799510 was just changing Safari 5-7 to use the typed array implementation.

It sounds like Chrome broke something in Canary. I opened an issue on the V8 tracker so we can fix this before this makes it farther in the release cycle. https://bugs.chromium.org/p/v8/issues/detail?id=4665

LinusU:
Hehe, that is some timing though :laughing: 

The output of this script in the above stated version of Chrome alerts `false` and `true`. In Chrome Stable, Version 47.0.2526.106 (64-bit), it alerts `true` and `true`.

```javascript
function typedArraySupportBroken () {
  function Bar () {}
  try {
    var arr = new Uint8Array(1)
    arr.foo = function () { return 42 }
    arr.constructor = Bar
    return arr.foo() === 42 && // typed array instances can be augmented
        arr.constructor === Bar && // constructor can be set
        typeof arr.subarray === 'function' && // chrome 9-10 lack `subarray`
        arr.subarray(1, 1).byteLength === 0 // ie10 has broken `subarray`
  } catch (e) {
    return false
[...]

PR #95, opened 2016-01-13
Fix Chrome 49

LinusU:
See #94, this may or may not be the best approach. At least it works :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -782,7 +782,9 @@ Buffer.prototype.slice = function slice (start, end) {
 
   var newBuf
   if (Buffer.TYPED_ARRAY_SUPPORT) {
+    this.__proto__ = Uint8Array.prototype
     newBuf = this.subarray(start, end)
+    this.__proto__ = Buffer.prototype
     newBuf.__proto__ = Buffer.prototype
   } else {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks for the PR, but https://github.com/feross/buffer/issues/94#issuecomment-172131066 says the bug is fixed in Chrome itself, and I can confirm that the tests all pass in Chrome 49.0.2623.0 canary (64-bit, Mac OS X). Closing unless you still think this is an issue, thanks! :)

LinusU:
Great!

littledan:
It'd be great if we could fix this bug in this repo as well. The ES2016 draft spec mandates the behavior that Chrome 49 tries out. Turns out it's not web-compatible yet (due to this particular library and possibly nothing else), but I think it's the most sensible semantics long-term. It'd be great if we could do something in this library so that the web can get the better TypedArray subclassing semantics within a few years.

LinusU:
It probably have to be discussed in Node.js as well, but I think that it's sensible to allow `new Buffer(buffer, start, end)` to create a new slice of another buffer. I'm all for making buffer as similar to typed arrays as possible.

Issue #86, opened 2015-12-18
Slicing buffers fails in latest Microsoft Edge


oskbor:
slicing buffers fails in Edge 25.10586.0.0
it boils down to this:
```
var buf = new Buffer(100);
var arr = new Uint8Array(100);
buf instanceof Uint8Array; // true
Uint8Array.prototype.subarray.call(arr, 0, 0).length; // 0
Uint8Array.prototype.subarray.call(buf, 0, 0).length; // 100 ! 
```
Would you say this is an issue with Edge or with this lib?

nolanlawson:
This sounds like an issue with Edge, especially if it's a regression from IE11. Should probably be filed on http://connect.microsoft.com/

feross:
This sounds like a regression in Edge from IE11. Can you file a bug report with Microsoft and post the link here?

feross:
Just added Edge to the test suite (https://github.com/feross/buffer/pull/87) and for what it's worth, the tests all pass.

oskbor:
I have tested it some more all tests pass in Edge, as you said. 
When i use Buffer from https://wzrd.in/standalone/buffer, I don't get these issues.
So it seems to be something in our tool chain that is causing this issue. Maybe some combination of Babel/browserify. I will continue down the rabbit hole, thanks for your assistance and for updating your tests to run on Edge

feross:
Thanks for updating this issue, @oskbor. If you end up discovering a bug, please do open a new issue, or comment in this one. Cheers!

oskbor:
Bug report posted to Microsoft
https://connect.microsoft.com/IE/feedbackdetail/view/2207846/buffer-uint8array-regression-in-latest-edge

feross:
I just marked that I've also experienced the bug on the issue.

oskbor:
Thanks!
Do you have any ideas on why my build might fail while the one on wzrd.in does not? I spent all day trying to find the reason, but ended up no wiser :confused:

feross:
@oskbor Can you open a new issue?

Sjors:
The test sweet uses Edge 13 which is ancient. This bug doesn't occur in Edge 20. It does in Edge 25. There are no VM's out there running Edge 25 as far as I know; I spent a ton of time getting my hands on a physical machine to reproduce this :-)

feross:
[...]

PR #97, opened 2016-01-14
Ensure ES2016 engines construct Uint8Array (not Buffer) from Buffer.p…

littledan:
…rototype.slice

In the ES2016 draft specification, TypedArray methods like
%TypedArray%.prototype.subarray() call out to a constructor for the result
based on the receiver. Ordinarily, the constructor is instance.constructor,
but subclasses can override this using the Symbol.species property on the
constructor.

Buffer.prototype.slice calls out to %TypedArray%.prototype.subarray, which
calls this calculated constructor with three arguments. The argument pattern
doesn't correspond to a constructor for Buffer, so without setting
Symbol.species appropriately, the wrong kind of result is created.

This patch sets Buffer[Symbol.species] to Uint8Array when appropriate, to
address the issue.

This address issue https://github.com/nodejs/node/issues/4701

### DIFF
--- a/index.js
+++ b/index.js
@@ -217,6 +217,11 @@ function fromJsonObject (that, object) {
 if (Buffer.TYPED_ARRAY_SUPPORT) {
   Buffer.prototype.__proto__ = Uint8Array.prototype
   Buffer.__proto__ = Uint8Array
+  if (typeof Symbol !== 'undefined' && Symbol.species &&
+      Buffer[Symbol.species] === Buffer) {
+    Object.defineProperty(Buffer, Symbol.species,
+                          { value: null, configurable: true })
+  }
 } else {
   // pre-set for values that may exist in the future
   Buffer.prototype.length = undefined


littledan:
This has been tested by running a simple test with Browserify in Chrome with @@species turned on, observing that slice generates something of the right length.

dcousens:
@feross where is the PR travis report? OOI

nolanlawson:
Thanks for the PR and excellent explanation! Could you please add a test to verify this functionality? Feel free to make the test exit early if the environment doesn't support `Symbol`.

As-is, Travis seems down, but I can guarantee that this PR will not pass in several browsers, because `Symbol` is not universally supported yet. So another good fix would be to have a test for `Symbol` (I recommend `typeof Symbol !== 'undefined'`.)

littledan:
I'm not sure if we need an extra test. I think browsers which encounter this issue would previously fail test/slice.js, and the test that I was using in development was basically a subset of this one.

littledan:
Yeah, on my test version of Chrome with spec-compliant Symbol.species, without this patch, 4 tests fail, and 82 tests pass. With the patch, all tests pass. I pushed a new version which feature-tests for Symbol.

index.js
```diff
@@ -217,6 +217,11 @@ function fromJsonObject (that, object) {
 if (Buffer.TYPED_ARRAY_SUPPORT) {
   Buffer.prototype.__proto__ = Uint8Array.prototype
   Buffer.__proto__ = Uint8Array
+  if (typeof global.Symbol !== 'undefined' && Symbol.species &&
```
nolanlawson: I would prefer `typeof Symbol` instead of `typeof global.Symbol`. Reason being: it might fail in odd environment like NW.js that have dual global objects.

oskbor:
We tried this pull request and it solved #86 for us. So we are hoping this lands soon!
br Oskar

index.js
littledan: What about browsers that don't support Symbol at all?

nolanlawson: `typeof Symbol` will return false for those browsers (and not throw an error)

littledan: Sorry for the delay; of course it does. Pushed a new version which is worded that way. What do you think?

oskbor:
@nolanlawson  are you happy with the latest changes from @littledan?

feross:
@littledan Thanks for all the help! Seriously... Quickly reverting the Canary change, sending this excellent PR, and explaining the issue so well. It's all very appreciated. You're awesome!

This is released as buffer 4.5.0. It's automatically included in browserify 13.0.0 (the latest version) because of semver.

littledan:
Thanks!

PR #98, opened 2016-01-16
package.json: use `files`

jscissr:
The following files will no longer be installed with `npm install buffer`:

- `.npmignore`
- `.travis.yml`
- `.zuul.yml`
- `test/**`

This makes the install faster.


nolanlawson:
I'm :+1:, but gonna leave it to @feross to merge. Any potential issues with not including the LICENSE or README with the project?

jscissr:
These are automatically included: https://docs.npmjs.com/files/package.json#files

nolanlawson:
Sweet!

nolanlawson:
Apparently the "main" is installed by default? https://twitter.com/ericclemmons/status/688429799404998656 If so, can we just do `"files": []`?

jscissr:
Yes, it works! :+1:

feross:
I personally don't think it's a good idea to ignore the `test/` folder. That's documentation for the module that will now be missing when you're offline and need to look inside `node_modules/buffer`.

Ignoring huge test files is a good idea, but ignoring a few text files won't make a meaningful difference. `npm install` is primarily slow because of latency and roundtrips, in my experience. I don't think this is the right tradeoff.

Previous discussions:

https://github.com/nodejs/readable-stream/pull/100
https://github.com/rvagg/through2/pull/36
https://github.com/substack/node-browserify/issues/1093
https://github.com/crypto-browserify/sha.js/issues/5
https://github.com/feross/string-to-stream/pull/1
https://github.com/beatgammit/base64-js/pull/16
https://github.com/mafintosh/csv-parser/pull/43

Issue #99, opened 2016-01-28
value out of bounds for floating point Infinity


substack:
Using browserify 13.0.0 which ships with buffer 4.3.1, I'm getting a RangeError when attempting to write a valid float value of -Infinity to a buffer:

```
$ echo "new Buffer(4096).writeFloatBE(-Infinity,0)" | browserify - | browser-run -b chrome
Uncaught RangeError: value is out of bounds
```

This seems to be an issue too with doubles:

```
$ echo "new Buffer(4096).writeDoubleBE(-Infinity,0)" | browserify - | browser-run -b chrome
Uncaught RangeError: value is out of bounds
```

Positive infinity yields the same error:

```
$ echo "new Buffer(4096).writeFloatBE(Infinity,0)" | browserify - | browser-run -b chrome
Uncaught RangeError: value is out of bounds
```

This works fine in node:

```
$ node -pe "new Buffer(4096).writeFloatBE(-Infinity,0)"
4
```

feross:
Released as 4.4.0.

PR #100, opened 2016-01-28
truncate toward infinity instead of raising an exception

substack:
fix for #99 to match node behavior to truncate towards ±Infinity instead of raising an exception

### DIFF
--- a/test/write_infinity.js
+++ b/test/write_infinity.js
@@ -1,0 +1,45 @@
+if (process.env.OBJECT_IMPL) global.TYPED_ARRAY_SUPPORT = false
+var B = require('../').Buffer
+var test = require('tape')
+
+test('write/read Infinity as a float', function (t) {
+  var buf = new B(4)
+  t.equal(buf.writeFloatBE(Infinity, 0), 4)
+  t.equal(buf.readFloatBE(0), Infinity)
+  t.end()
+})
+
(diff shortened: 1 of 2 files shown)


dcousens:
ACK,  is this matched with all versions of node or just recently?

test/write_infinity.js
```diff
@@ -0,0 +1,45 @@
+if (process.env.OBJECT_IMPL) global.TYPED_ARRAY_SUPPORT = false
+var B = require('../').Buffer
+var test = require('tape')
```
dcousens: Should this be `var t`?

dcousens: nvm

feross:
@dcousens If this matches behavior in the latest node, it's fine to merge. We track the latest version.

feross:
Running the full browser test suite locally...

substack:
Old versions of node worked this way too:

```
$ node -v
v0.10.41
$ node -pe "b=Buffer(4); b.writeFloatBE(Infinity,0); b.readFloatBE(0)"
Infinity
$ node -pe "b=Buffer(4); b.writeFloatBE(4e38,0); b.readFloatBE(0)"
Infinity
```

feross:
Released as 4.4.0.

PR #112, opened 2016-04-20
Update buffer for node.js v6

### LINKED ISSUE #111: allocUnsafe/alloc/from
Any plans about supporting new static functions? Ref: https://github.com/nodejs/node/pull/4682/

feross:
So, this took way longer that I expected. Worked on it for two days. The node.js buffer test suite has improved quite a bit, so there were a lot of bugs to fix!

This PR brings buffer up-to-date with node.js commit 6275249 on Apr 15, 2016. (https://github.com/nodejs/node/blob/e38bade82801645d8cc1010e3f4e9826052244cb/lib/buffer.js)

### Added

- Buffer.from
- Buffer.alloc
- Buffer.allocUnsafe
- Buffer.allocUnsafeSlow
- Buffer.prototype.swap16
- Buffer.prototype.swap32
- Buffer.prototype.includes
- expose constant `kMaxLength`

### Changed

The following functions now take additional arguments:

- Buffer.prototype.compare(target, start, end, thisStart, thisEnd)
- Buffer.prototype.indexOf(val, byteOffset, encoding)
- Buffer.prototype.fill(val, start, end, encoding)

Also, too many bug fixes to count in order to bring the package inline with the latest node.js test suite.

### Removed

- Removed `.parent` properties. They're not used for anything, and the test suite doesn't check for them.

Also, node.js uses ES6 freely in the test suite now, which is really frustrating for projects like this and `readable-stream` which are also consumed in the browser. It makes it impossible to use the tests in browsers without support for arrow functions, for example. Babel isn't an option either since it has lots of complaints about the code, like top-level return, as one example.

So, we only run the node tests in the newest browsers, now. This technically reduces the test coverage for older browsers, but we don't really have another choice if we want to keep the test suite in sync. Old browsers back to IE8 are still supported, but they will run our own test suite only, not the node.js one. Onward and forward!

Fixes #111.


fijiwebdesign:
Is it not possible to transpile the Node.js ES6 tests to ES5 or ES3 and run the tests in the browser?

feross:
@fijiwebdesign Tried that and ran into babel issues. It doesn't like top-level `return`s, even though it's perfectly fine in commonJS.

fijiwebdesign:
Wouldn't browserify fix that by wrapping it in a function? So transpile to ES3 then browserify? Just wondering.

feross:
Yeah, but Babel thinks every file it sees is an ES6 module.

feross:
Released as 4.6.0.

PR #117, opened 2016-05-10
Call correct set method on Buffer.copy()

felix:
This is a fix from the 3.6.0 tag (no branch).

Current stable Webpack (v1.13.0) depends on node-libs-browser <= 0.6.0 which in turn depends on buffer ^3.0.3 which pulls in v3.6.0 which still has this bug.


dcousens:
For reference,  this PR was rejected for the reasons set out by @feross in https://github.com/feross/buffer/issues/118#issuecomment-219594693

Issue #121, opened 2016-06-18
IE10: Object doesn't support property or method 'fill'


kawanet:
It looks the current version 4.6.0 does not work on IE10.

- IE9 + buffer 4.6.0 - OK
- IE10 + buffer 4.6.0 - NG
- IE11 + buffer 4.6.0 - OK
- IE10 + buffer 4.1.0/4.5.1 - NG
- IE10 + buffer 3.6.0/4.0.0 - OK

kawanet:
How to reproduce the problem:

```sh
echo 'document.write([].join.call(Buffer(1).fill(123)));' > src.js 
browserify src.js -o out.js
echo '<script src="./out.js"></script>' > index.html 
```

- `123` on document body: OK
- `Object doesn't support property or method 'fill'` error on console: NG

kawanet:
This does not work on IE10:

```js
document.write([].join.call(Buffer(1).fill(123)));
```

This does work on IE10:

```js
document.write([].join.call(Buffer.prototype.fill.call(Buffer(1), 123)));
```

kawanet:
I got the reason.

The "IE10" I mentioned above is an IE10 mode on the IE11 switched with the built-in developer tools.

It works on the original IE10.

- IE10-on-IE11 has TYPED_ARRAY_SUPPORT = true
- The original IE10 has TYPED_ARRAY_SUPPORT = false

nolanlawson:
So if I understand correctly, this only fails when you're using IE11 but
trying to emulate older IEs via the doctype switch?

kawanet:
@nolanlawson Right.

I found this when testing another application for IE10 but on IE11.

![ie10](https://cloud.githubusercontent.com/assets/9765/16181030/9af35a06-36cf-11e6-9d8c-867e693067e8.png)

Any developers who want to test on the environment would face the problem.
It seems that it has happened since 12a3318 of v4.1.0.
The patch 23548e7 respects the new way of augmentation using `__proto__`.

feross:
Thanks for digging into this. I just confirmed what you found. This is definitely a developers-only edge case, but you're right to point out that the code that's checking for typed array augmentation should just check if `__proto__` can be set, since that's what we do now.

Released as 4.6.1.

andreamancuso:
Hello folks, I have come across this issue using native IE10 (on VirtualBox Win7). I suppose I'm just very unlucky?

feross:
[...]

PR #122, opened 2016-06-18
re #121: IE10-on-IE11 fails the augmentation

kawanet:
This patch fixes the problem found on IE10 mode of IE11 developer tools.
It follows the new way to augment Uint8Array instances on typedArraySupport method.


feross:
Released as 4.6.1

Issue #114, opened 2016-04-28
Add missing .lastIndexOf method


feross:
Node.js documentation for .lastIndexOf: https://nodejs.org/api/buffer.html#buffer_buf_lastindexof_value_byteoffset_encoding

> Identical to Buffer#indexOf(), but searches the Buffer from back to front instead of front to back. Returns the starting index position of value in Buffer or -1 if the Buffer does not contain value. The value can be a String, Buffer or Number. Strings are by default interpreted as UTF8. If byteOffset is provided, will return the last match that begins at or before byteOffset.

@dcposch's PR: https://github.com/nodejs/node/pull/4846

feross:
@dcposch -- any interest in sending a PR to implement this? ;)

dcposch:
lol k

feross:
Yo @dcposch, no pressure, but this issue is still open if you want it 😉

dcposch:
sure, on it

feross:
Awesome! Fixed by #127

PR #126, opened 2016-08-08
Update node.js tests and fix the bugs it caught

feross:
There have been lots of changes to the node.js buffer tests since we last synced them. This PR brings our tests in line with what's in Node.js core. These also helped us catch several bugs in our implementation:

- Fix: treat numbers as unsigned 8-bit for .indexOf and .includes
- Fix: buffer should only be as long as the valid part of a hex string
- Add 'latin1' encoding (alias for 'binary')
- Remove deprecated 'raw' and 'raws' encoding

There are a few tests that we're still failing (but this has been the case for a while): 

- Handling invalid base64 better -- truncating the buffer at the point of the invalid data
- missing the .lastIndexOf (see #114)

The asserts that relate to the above issues remain commented out. Additionally, a stub implementation of .lastIndexOf has been added that just throws an exception instead of letting the Uint8Array implementation's .lastIndexOf be exposed, since that one has a different interface.

Also, pulling in these new tests uncovered a bug in Node.js's `assert` package. See: https://github.com/nodejs/node/issues/8001 So there's a dirty hack to workaround that in `test/node/test-buffer-badhex.js` until the issue is fixed in a shipping version of Node.js v6.

### DIFF
--- a/index.js
+++ b/index.js
@@ -228,7 +228,12 @@ function fromString (that, string, encoding) {
   var length = byteLength(string, encoding) | 0
   that = createBuffer(that, length)
 
-  that.write(string, encoding)
+  var actual = that.write(string, encoding)
+
+  if (actual !== length) {
+    that = that.slice(0, actual)
+  }
+
   return that
 }
 
--- a/index.js
+++ b/index.js
@@ -351,6 +351,7 @@ Buffer.isEncoding = function isEncoding (encoding) {
     case 'utf8':
     case 'utf-8':
     case 'ascii':
+    case 'latin1':
     case 'binary':
     case 'base64':
     case 'ucs2':
@@ -413,6 +414,7 @@ function byteLength (string, encoding) {
   for (;;) {
     switch (encoding) {
       case 'ascii':
+      case 'latin1':
       case 'binary':
         return len
       case 'utf8':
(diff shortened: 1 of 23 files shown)


feross:
Can I get a review, please? @jessetane @dcousens @nolanlawson Thanks in advance guys!

index.js
```diff
@@ -228,7 +228,12 @@ function fromString (that, string, encoding) {
   var length = byteLength(string, encoding) | 0
   that = createBuffer(that, length)
 
-  that.write(string, encoding)
+  var actual = that.write(string, encoding)
+
+  if (actual !== length) {
```
dcposch: When would `write` only write part of the string?

dcposch:
LGTM

index.js
feross: This happens when writing a hex string that contains invalid characters. For example, writing `abcxxabc` will only write `abc`. Node ignores everything after the first invalid character.

You can test it:

```js
Buffer.from('abcdxxabcd', 'hex').toString('hex')
```

feross: I'll add a comment.

dcousens:
ACK up to 50b0526

feross:
Released as 4.8.0.

PR #128, opened 2016-08-08
Fix: Add missing "typeof"

feross:
Without this, we never go down the faster native indexOf code-path. So
this wasn't causing bugs in user code, it was just probably slower than
necessary.


feross:
Released as 4.9.0.

PR #132, opened 2016-08-18
sync tests; throw on negative .allocUnsafe() argument; remove hacks

feross:
- sync node.js tests to master (node v6.4.0)
- throw on negative .allocUnsafe() argument (to match new node v6.4.0 behavior)
- remove assert hack (Bug in node.js core was fixed in node v6.4.0!)


feross:
Released as 4.9.1

Issue #133, opened 2016-09-13
UTF8 decoding differs from Node


dcousens:
### Code
``` js
var salt = 'mnemonicメートルガバヴァぱばぐゞちぢ十人十色'
var buffer = new Buffer(salt, 'utf8')

console.log(buffer.toString('base64'))
```

### Results

**node v6.5.0**
``` js
bW5lbW9uaWPjg6Hjg7zjg4jjg6vjgqvjgpnjg4/jgpnjgqbjgpnjgqHjga/jgprjga/jgpnjgY/jgpnjgp3jgpnjgaHjgaHjgpnljYHkurrljYHoibI=
```

**Chromium 53.0.2785.101**
``` js
bW5lbW9uaWPDo8aSwqHDo8aSwrzDo8aSy4bDo8aSwqvDo+KAmsKrw6PigJrihKLDo8aSwo/Do+KAmuKEosOj4oCawqbDo+KAmuKEosOj4oCawqHDo8KBwq/Do+KAmsWhw6PCgcKvw6PigJrihKLDo8KBwo/Do+KAmuKEosOj4oCawp3Do+KAmuKEosOjwoHCocOjwoHCocOj4oCa4oSiw6XCjcKBw6TCusK6w6XCjcKBw6jigLDCsg==
```

jessetane:
@dcousens is your document's character encoding set to utf-8?

`<meta charset=utf-8>`

dcousens:
@jessetane nailed it.

How do we help make that obvious... can we perhaps inform users?

jessetane:
Interesting idea, something like `if ('ト'.length !== 1) throw new Error('encoding error')`?

dcousens:
When the effect is so subtle,  I think such an error would be greatly appreciated by most users.

jessetane:
For what it's worth it looks like Firefox actually already does warn if your document's encoding is unspecified.

feross:
Wow, I had no idea that the html document encoding could affect javascript strings! I guess that's just because I always use a utf8 document encoding.

feross:
Sent a PR to fix this: https://github.com/feross/buffer/pull/134

jessetane:
Just to be precise about why this happened: the browser can only guess what encoding a source file is in unless you specify it via meta tag or http header. Of course everyone stores text files as utf-8 these days, but for backward compatibility the browser probably assumed iso-8859-1 and so OP just experienced this mismatch.

feross:
[...]

PR #134, opened 2016-09-14
Add warning when document encoding may create issues

feross:
Fixes: https://github.com/feross/buffer/issues/133

Print the following warning if the document encoding may create issues
for `buffer`:

> The character encoding of the HTML document was not declared as UTF-8.
Therefore, the `buffer` package may behave incorrectly in some
situations. Consider adding `<meta charset="utf-8">` to the `<head>`
section of the HTML document.

### DIFF
--- a/index.js
+++ b/index.js
@@ -125,6 +125,16 @@ if (Buffer.TYPED_ARRAY_SUPPORT) {
     })
   }
 }
+
+if ('ト'.length !== 1) {
+  console.warn(
+    'The character encoding of the HTML document was not declared as UTF-8. ' +
+    'Therefore, the `buffer` package may behave incorrectly in some situations. ' +
+    'Consider adding <meta charset="utf-8"> to the <head> section of the HTML ' +
+    'document.'
+  )
+}
+
 Buffer.poolSize = 8192 // not used by this implementation
 
 // TODO: Legacy, not needed anymore. Remove in next major version.
--- a/index.js
+++ b/index.js
@@ -113,6 +113,18 @@ function Buffer (arg, encodingOrOffset, length) {
   return from(this, arg, encodingOrOffset, length)
 }
 
+if (Buffer.TYPED_ARRAY_SUPPORT) {
+  Buffer.prototype.__proto__ = Uint8Array.prototype
+  Buffer.__proto__ = Uint8Array
+  if (typeof Symbol !== 'undefined' && Symbol.species &&
+      Buffer[Symbol.species] === Buffer) {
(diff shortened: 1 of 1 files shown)


jessetane:
It's not so much that Buffer requires the document's character encoding to be utf8, just that the source file encoding matches the document's.

feross:
@jessetane Can you suggest an improved warning message?

jessetane:
This might be tricky, since I'm not sure it's possible to reliably detect arbitrary encoding mismatches, the test here just happens to work for this particular case... I'll think about it.

jessetane:
Ok so, there is [`document.characterSet`](https://dom.spec.whatwg.org/#dom-document-characterset), which on my machine at least defaults to `'windows-1252'` (not exactly latin-1 after all!). It's unclear to me though whether it should be Buffer's job to look at this though?

feross:
The goal of this package is to mimic the node.js buffer API exactly. So if the encoding is set in a way that will make it work differently than node.js then I think it's fair to warn the user.

The question for me is: does the encoding *have* to be UTF-8, or are there other encodings that are okay? i.e. does the `'ト'.length !== 1` test pass for other encodings and is that sufficient to guarantee that we'll get correct behavior?

At the very least, it seems that if `'ト'.length !== 1` is true, then we have a problem and should warn the user, right?

jessetane:
Buffer will function properly as long as no characters that compose its source are misinterpreted due to an encoding mismatch. In #133 the issue was related to the fact that the string literal was misinterpreted, not Buffer itself.  Buffer can likely be served safely in a wide range of formats including ones that don't even support Unicode, such as `'windows-1252'`.

I think in general the rule is just that you just have to know what format your data is in, and if you send that data to anyone else, you must be explicit about it - no exceptions!

feross:
> Buffer can likely be served safely in a wide range of formats including ones that don't even support Unicode, such as 'windows-1252'.

Gotcha. Then it sounds like this check doesn't belong inside this package.

jessetane:
Yeah, it should be the browser's job, since only it can know for sure that it took a guess at the encoding. Here's what Firefox says if you miss the meta tag / http header:
> The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.

dcousens:
If you are using `utf-8` as your encoding, and your document isn't `utf-8`... you're going to probably have a bad time.
I still think the warning has merit, personally.  But I also agree that my browser let me down.

PR #137, opened 2016-09-23
Remove isarray dep.

jdalton:
Hi @feross!

The [isarray](https://www.npmjs.com/package/isarray) package [is being deprecated](https://github.com/juliangruber/isarray/issues/10) since IE < 11 is no longer supported by MS & IE 8 was the last browser that needed it. For the fringes, folks tend to use something like [es-shim](https://github.com/es-shims).

The `Array.isArray` method is supported by all Node versions, Rhino, iOS, Android, and headless browser enviros (like PhantomJS v1, SlimerJS).

With that in mind this PR removes it from `buffer`.

*Note: I don't think test fails are related to this change.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,12 +21,12 @@ instance methods, and class methods that are supported.
 
 ## features
 
-- Manipulate binary data like a boss, in all browsers -- even IE6!
+- Manipulate binary data like a boss, in all browsers!
 - Super fast. Backed by Typed Arrays (`Uint8Array`/`ArrayBuffer`, not `Object`)
 - Extremely small bundle size (**5.04KB minified + gzipped**, 35.5KB with comments)
-- Excellent browser support (IE 6+, Chrome 4+, Firefox 3+, Safari 5.1+, Opera 11+, iOS, etc.)
+- Excellent browser support (IE 11, Edge, Chrome, Firefox, Safari 5.1+, Opera, iOS, etc.)
 - Preserves Node API exactly, with one minor difference (see below)
-- Square-bracket `buf[4]` notation works, even in old browsers like IE6!
+- Square-bracket `buf[4]` notation works!
 - Does not modify any browser prototypes or put anything on `window`
 - Comprehensive test suite (including all buffer tests from node.js core)
 
--- a/bin/zuul-es5.yml
+++ b/bin/zuul-es5.yml
@@ -5,7 +5,7 @@ browsers:
   - name: safari
     version: latest
   - name: ie
-    version: 8..latest
+    version: 11..latest
   - name: microsoftedge
     version: 13..latest
   - name: android
--- a/index.js
+++ b/index.js
@@ -10,7 +10,6 @@
 
 var base64 = require('base64-js')
 var ieee754 = require('ieee754')
-var isArray = require('isarray')
 
 exports.Buffer = Buffer
 exports.SlowBuffer = SlowBuffer
@@ -302,7 +301,7 @@ function fromObject (that, obj) {
       return fromArrayLike(that, obj)
     }
 
-    if (obj.type === 'Buffer' && isArray(obj.data)) {
+    if (obj.type === 'Buffer' && Array.isArray(obj.data)) {
       return fromArrayLike(that, obj.data)
     }
   }
@@ -374,7 +373,7 @@ Buffer.isEncoding = function isEncoding (encoding) {
 }
 
 Buffer.concat = function concat (list, length) {
-  if (!isArray(list)) {
+  if (!Array.isArray(list)) {
     throw new TypeError('"list" argument must be an Array of Buffers')
   }
 
--- a/package.json
+++ b/package.json
@@ -16,8 +16,7 @@
   ],
   "dependencies": {
     "base64-js": "^1.0.2",
-    "ieee754": "^1.1.4",
-    "isarray": "^1.0.0"
+    "ieee754": "^1.1.4"
   },
   "devDependencies": {
     "benchmark": "^2.0.0",


jdalton:
I don't think test fails are related to this change. BTW `npm test` fails in Node v6 due to a readonly assignment to `Buffer.from =`.

kgryte:
One of the selling points is that this module supports everything from IE6 and up. I see no reason to remove this dep.

jdalton:
IE6 has less than 0.2% market share on a dead OS. The `buffer` package ends up in a lot of other deps so while it's a nice academic exercise to expand support to ancient enviros it's being shipped to plenty of folks with little to no benefit.

kgryte:
Backward compatibility is not an academic exercise. I am one of those 0.2% who has no other choice but to use a dead ~16yr old OS. Our image acquisition software cannot be updated to a newer OS due to driver compatibility issues. Until usage is 0% and will forever remain 0%, backward compatibility matters. People at the margins are still people, and their concerns should not be dismissed so readily in the name of modernity.

jdalton:
To help move this along the next version of Lodash [will poison the API](https://github.com/lodash/lodash/commit/ab904a10473c1b703af22fe387fc0c34bab39c54).

feross:
@jdalton I understand where you're coming from -- pruning unneeded dependencies is one of my favorite pastimes. It's something that I wish more node developers took seriously, especially for packages that just run on the server-side, a much simpler and predictable environment. Reducing install time, decreasing security attack surface area, increasing site performance, and simplifying codebases are all great benefits.

So I'm definitely on your side in the fight for "dependency hygiene". But this PR drops support for IE8 a bit prematurely, IMO.

IE 6-8 together account for .77% of the US market (.08% IE6 + .07% IE7 + .62% IE8). There are 269.35 million US internet users, so this PR leaves behind **2.07 million people**. And that's just in the US.

If the complexity or file size savings were meaningful, I could justify it. But this only saves 3 lines of code. Is this really the right tradeoff?

feross:
> @jdalton To help move this along the next version of Lodash will poison the API.

If I understand this correctly -- you're saying that anyone who uses `lodash` on IE8 will get an intentionally broken `Array.isArray` going forward? And that broken `Array.isArray` will also break our usage of the `isarray` package.

Why oh why? Is this necessary?

mxstbr:
> you're saying that going forward anyone who uses lodash on IE8 will get an intentionally broken Array.isArray?

If I understand correctly that's totally not the case. If you take a look at the diff linked, this is what the code is:

```JS
Array.isArray = Array.isArray || (Array.isArray = function() {
  throw new Error('This method is not supported. Try https://github.com/es-shims.');
});
```

which means it'll throw a helpful error when it's undefined, instead of just throwing "Undefined is not a function".

feross:
@mxstbr He called it "poisoning the API" for a reason. It prevents anyone else from easily detecting that `Array.isArray` is missing. So packages like `isarray` will use `lodash`'s "helpful" implementation:

```js
var toString = {}.toString;
module.exports = Array.isArray || function (arr) {
  return toString.call(arr) == '[object Array]';
}
```
https://github.com/juliangruber/isarray/blob/master/index.js

jdalton:
@feross 
> If I understand this correctly -- you're saying that anyone who uses lodash on IE8 will get an intentionally broken `Array.isArray` going forward?

It's more of a show of intent since currently Lodash v4 doesn't work on an unshimmed IE8 in the first place.

feross:
> It's more of a show of intent since currently Lodash v4 doesn't work on an unshimmed IE8 in the first place.

Hmm, I see. So, you're saying that lodash users already need to use `es-shim` anyway to make it work in IE8. I get that.

I still feel like doing a hostile takeover of the `Array.isArray` and breaking other people's packages (who do include a shim) isn't necessary.

If you want to help unshimmed IE8 users, you can just print them a helpful console warning, or throw an exception, as soon as `lodash` is required.

jdalton:
It's my hope that by a more popular library taking the plunge it helps remove any hangups you may have. If folks complain then I'll revert but I don't anticipate many issues.

feross:
> it helps remove any hangups you may have

I think it's fine to drop support for older browsers in a major version of `buffer`. No hangups. That's valid semver. It's just that, as it stands, this PR is a net loss because it removes browser support and adds nothing else.

If we start using `indexOf` instead of handwritten for-loops to reduce code complexity and file size, and other IE9+ niceties, then we can justify dropping the older users in the next version. (Of course, people can just continue using their current version of browserify if they want older browser support). If you or someone else wants to take a stab at that, I'm happy to merge it.

jdalton:
> I think it's fine to drop support for older browsers in a major version of buffer. No hangups. That's valid semver.

Sweet! Would you dig creating a v5 branch to apply PRs to?

> If you or someone else wants to take a stab at that, I'm happy to merge it.

Sure thing!

feross:
@jdalton Okay, I created a v5 branch. Want to re-send this PR?

jdalton:
@feross Thanks! Github [introduced a feature](https://twitter.com/github/status/765259651772448769) so there's no need to close and re-send I can just change the target of this one.

jdalton:
If you re-open it should be pointing to the v5 branch.

README.md
```diff
@@ -21,12 +21,12 @@ instance methods, and class methods that are supported.
 
 ## features
 
-- Manipulate binary data like a boss, in all browsers -- even IE6!
+- Manipulate binary data like a boss, in all browsers!
 - Super fast. Backed by Typed Arrays (`Uint8Array`/`ArrayBuffer`, not `Object`)
 - Extremely small bundle size (**5.04KB minified + gzipped**, 35.5KB with comments)
-- Excellent browser support (IE 6+, Chrome 4+, Firefox 3+, Safari 5.1+, Opera 11+, iOS, etc.)
+- Excellent browser support (IE, Edge, Chrome, Firefox, Safari, Opera, iOS, etc.)
```
feross: Thanks for adding Edge to the list here, but can you keep the browser versions for all but Edge, Chrome, and Firefox (the evergreen browsers).

jdalton:
:metal:

README.md
jdalton: No prob!

Would you like to list the supported IE version by MS or technically by the code?
(it means either IE9 or IE11).

Do you consider Opera and Safari evergreen?

feross: Let's list IE11. If we also drop IE9 and IE10 while we're at it, we can kill off the Object implementation of Buffer entirely, and just leave the Uint8Array implementation which would reduce file size significantly.

IE10 technically has Uint8Array but it's buggy, so it uses the Object implementation.

Opera is evergreen, but Safari is not.

PR #138, opened 2016-09-24
Remove Object implementation

feross:
Reduces a bit of complexity, and around 4KB (unzipped). Not as large of a difference as I'd hoped.

Interested persons should send PRs to the v5 branch that could further simplify the code, with the assumption that IE11+ is required.

cc @jdalton

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,18 +15,20 @@ exports.Buffer = Buffer
 exports.SlowBuffer = SlowBuffer
 exports.INSPECT_MAX_BYTES = 50
 
+var K_MAX_LENGTH = 0x7fffffff
+exports.kMaxLength = K_MAX_LENGTH
+
 /**
- * If `Buffer.TYPED_ARRAY_SUPPORT`:
+ * If `TYPED_ARRAY_SUPPORT`:
  *   === true    Use Uint8Array implementation (fastest)
- *   === false   Use Object implementation (most compatible, even IE6)
+ *   === false   Print warning and recommend using `buffer` v4.x which has an Object
+ *               implementation (most compatible, even IE6)
  *
  * Browsers that support typed arrays are IE 10+, Firefox 4+, Chrome 7+, Safari 5.1+,
  * Opera 11.6+, iOS 4.2+.
  *
- * Due to various browser bugs, sometimes the Object implementation will be used even
- * when the browser supports typed arrays.
- *
- * Note:
+ * Sometimes we report that the browser does not support typed arrays, if there
+ * exist extreme bugs in the implemetation. For example:
  *
  *   - Firefox 4-29 lacks support for adding new properties to `Uint8Array` instances,
  *     See: https://bugzilla.mozilla.org/show_bug.cgi?id=695438.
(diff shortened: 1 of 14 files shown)


index.js
```diff
@@ -112,26 +91,20 @@ function Buffer (arg, encodingOrOffset, length) {
...
-  arr.__proto__ = Buffer.prototype
-  return arr
+// Fix subarray() in ES2016. See: https://github.com/feross/buffer/pull/97
+if (typeof Symbol !== 'undefined' && Symbol.species &&
+    Buffer[Symbol.species] === Buffer) {
+  Object.defineProperty(Buffer, Symbol.species, {
+    value: null,
+    configurable: true
+  })
 }
```
jdalton: Should this contain the other `false` descriptors  to avoid accidental overrides if folks modify something like `Object.prototype.enumerable=true`?

feross: @littledan wrote this code, maybe we can get his thoughts?

littledan: Including explicit `false` sounds fine to me. I wasn't really considering such exotic modifications... this code isn't bullet-proof against all possible weird pieces of code someone may write, as it does leave the property configurable.

index.js
```diff
@@ -1300,79 +1251,47 @@ Buffer.prototype.writeUInt8 = function writeUInt8 (value, offset, noAssert) {
...
-function objectWriteUInt16 (buf, value, offset, littleEndian) {
-  if (value < 0) value = 0xffff + value + 1
-  for (var i = 0, j = Math.min(buf.length - offset, 2); i < j; ++i) {
-    buf[offset + i] = (value & (0xff << (8 * (littleEndian ? i : 1 - i)))) >>>
-      (littleEndian ? i : 1 - i) * 8
-  }
-}
-
 Buffer.prototype.writeUInt16LE = function writeUInt16LE (value, offset, noAssert) {
   value = +value
```
jdalton: I noticed Node is converting `offset` to a [unsigned int](https://github.com/nodejs/node/blob/v6.6.0/lib/buffer.js#L1065) and this is converting it to a signed one (`>>>` vs. `|`).

index.js
feross: @littledan Cool, thanks. @jdalton Want to send a PR that adds whatever properties you think should be here?

jdalton: Oh rock! I think this'll be my PR on a PR!

index.js
```diff
@@ -35,18 +37,14 @@ exports.INSPECT_MAX_BYTES = 50
...
- */
-exports.kMaxLength = kMaxLength()
+if (!Buffer.TYPED_ARRAY_SUPPORT) {
+  console.error(
+    'This browser lacks typed array (Uint8Array) support which is required by ' +
+    '`buffer` v5.x. Use v4.x if you require old browser support.')
+}
 
 function typedArraySupport () {
   try {
```
feross: Are you saying that since IE10 lacks __proto__, we don't need to test for broken subarray?

index.js
feross: Nice, I did that a while ago. It was too fun 👍

index.js
feross: Good catch

index.js
jdalton: `__proto__` looks critical to this shim, since IE 10 lacks support for a critical bit and since the bar is raised to IE 11 the IE 10 code paths can be dropped.

index.js
feross: Updated PR to fix this

index.js
feross: Yeah, without the object implementation, this just flat-out won't work in IE10 anymore.

index.js
```diff
@@ -15,73 +15,49 @@ exports.Buffer = Buffer
...
   }
 }
 
-function kMaxLength () {
-  return Buffer.TYPED_ARRAY_SUPPORT
-    ? 0x7fffffff
-    : 0x3fffffff
-}
-
 function createBuffer (that, length) {
```
jdalton: The `that` param of `createBuffer` is not used (it's overwritten without prior use) so it can be dropped and the signature simplified to `createBuffer (length)`

feross: Good point, this actually lets us remove lots of parameters.

feross:
Okay, this PR is getting too large. I'll merge and release v5.0.0. More improvements along these same lines are welcome.

feross:
buffer v5.0.0 is released.

jdalton:
👏  and sauce is all green!

PR #139, opened 2016-09-24
Update dot files.

jdalton:
This updates the .gitignore to avoid the `node_modules` folder and adds a .gitattirbutes for consistent line ending handling across platforms.


jdalton:
Thank you!

PR #142, opened 2016-09-27
Style improvements

feross:


index.js
```diff
@@ -156,7 +156,7 @@ function alloc (size, fill, encoding) {
  * Creates a new filled Buffer instance.
  * alloc(size[, fill[, encoding]])
  **/
-Buffer.alloc = function (size, fill, encoding) {
+Buffer.alloc = function alloc (size, fill, encoding) {
   return alloc(size, fill, encoding)
```
dcousens: won't this/why isn't this circular? (same question for the rest)

feross: You're right - they are. Tests don't pass. Sorry, I'm getting tired here 😵

PR #143, opened 2016-09-27
Remove old browser note. [ci skip]

jdalton:
I noticed the readme still had an old browser note (those that don't support typed arrays).


feross:
Thanks, LGTM!

PR #148, opened 2016-11-03
adventure-yunfei fix: wrong overwritting of not writable Unit8Array.from

adventure-yunfei:
`Uint8Array` defined an unwritable `from` property, which causes overwritting operation after changing `Buffer.__proto__` fails and break script execution.

e.g. failing to assign ".from" property in **Chrome**:
image

When use `buffer` with webpack, everything after `Buffer.from = ...` just gone.

And e.g. failing to require `buffer` in **node@4**
image

We should move "Buffer.from" assignment forward to avoid this error.

Note that this may not be reproducible on all browser. In some browsers (or node@6) `Uint8Array` doesn't have 'from', or defines a writable 'from' property.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,9 +84,6 @@ function Buffer (arg, encodingOrOffset, length) {
   return from(arg, encodingOrOffset, length)
 }
 
-Buffer.prototype.__proto__ = Uint8Array.prototype
-Buffer.__proto__ = Uint8Array
-
 // Fix subarray() in ES2016. See: https://github.com/feross/buffer/pull/97
 if (typeof Symbol !== 'undefined' && Symbol.species &&
     Buffer[Symbol.species] === Buffer) {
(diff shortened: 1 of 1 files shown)


feross:
What version of Chrome are you using? I've never seen this error.

feross:
I suppose this can't do any harm, so why not?

feross:
Released as 5.0.2

adventure-yunfei:
@feross  I've printed out chrome version (which is 50.0.2661.102), and node version (which is 4.4.4). On these platforms, `Object.getOwnPropertyDescriptor(Uint8Array.__proto__, 'from')` will get a non-writable descriptor. 

(And I've also tested latest chrome@55.0 and node@6.9.1 which shows it's writable instead.)

feross:
@adventure-yunfei Sounds good. Thanks for testing!

PR #149, opened 2016-11-07
Fix bug with .fill bytes > 127 from other encoding

emilbayes:


dcousens:
LGTM,  but not sure why it was that way before

emilbayes:
@dcousens Node.js delegates `.fill` to a native method. Maybe it was a brain fart? As far as I can see there are no tests covering bytes > 127, but I did't read all the Node tests either so my knowledge is only superficial

test/base64.js
```diff
@@ -44,3 +44,12 @@ test('base64: invalid non-alphanumeric characters -- should be stripped', functi
   )
   t.end()
 })
+
+test('base64: high byte', function (t) {
+  var highByte = Buffer.from([128])
```
feross: This should be `B.from`, right? None of the other tests reference `Buffer`.

emilbayes: Oh woops

PR #153, opened 2017-02-09
Small fixes

feross:
See individual commit messages.

### DIFF
--- a/bin/download-node-tests.js
+++ b/bin/download-node-tests.js
@@ -67,7 +67,7 @@ function testfixer (filename) {
       // require buffer explicitly
       var preamble = 'var Buffer = require(\'../../\').Buffer;\n'
       if (/use strict/.test(line)) line += '\n' + preamble
-      else line + preamble + '\n' + line
+      else line += preamble + '\n' + line
       firstline = false
     }
 
--- a/bin/download-node-tests.js
+++ b/bin/download-node-tests.js
@@ -65,7 +65,7 @@ function testfixer (filename) {
 
     if (firstline) {
       // require buffer explicitly
-      var preamble = 'var Buffer = require(\'../../\').Buffer;\n'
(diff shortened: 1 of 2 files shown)


bin/download-node-tests.js
```diff
@@ -65,9 +65,9 @@ function testfixer (filename) {
 
     if (firstline) {
       // require buffer explicitly
-      var preamble = 'var Buffer = require(\'../../\').Buffer;\n'
+      var preamble = 'var Buffer = require(\'../../\').Buffer;'
       if (/use strict/.test(line)) line += '\n' + preamble
-      else line + preamble + '\n' + line
+      else line += preamble + '\n' + line
```
dcousens: :laughing:

feross: Yeah, fortunately this didn't have any effect because all node core files have 'use strict' at the top. Heh.

feross:
> What was the originally reasoning not to do this?

Explanation for this used to be in the readme. It was a limitation of the approach I took to get Buffer to be backed by a `Uint8Array`.

https://github.com/feross/buffer/blob/580e30376ab273d7314243a9a98bf9c3a600b93a/README.md#important-always-use-bufferisbuffer-instead-of-instanceof-buffer

In short, we couldn't properly subclass the Uint8Array until recently, so Buffers were actually just Uint8Array instances with a bunch of methods tacked on. Yes, every "buffer" actually had it's own set of properties for all the buffer methods.

You can see it here: https://github.com/feross/buffer/blob/580e30376ab273d7314243a9a98bf9c3a600b93a/index.js#L1081-L1142

feross:
Released as 5.0.4.

dcousens:
Thanks for the explanation,  it had just been a while so I figured I'd ask to be sure :+1:

Issue #154, opened 2017-02-09
5.0.4 has broken change


legokichi:
```js
// fileA.js
const Buffer = require("buffer/").Buffer;
module.exports.A = new Buffer(10);
```

```js
// fileB.js
const Buffer = require("buffer/").Buffer;
module.exports.B = new Buffer(10);
```

```js
// fileC.js
const {A} = require("./fileA");
const {B} = require("./fileB");
console.log(Buffer.concat([A, B]));
```

in 5.0.3
```
$ browserify fileC.js > fileD.js | node
<Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>
```

in 5.0.4

```
$ browserify fileC.js > fileD.js | node
[stdin]:472
      throw new TypeError('"list" argument must be an Array of Buffers')
      ^

TypeError: "list" argument must be an Array of Buffers
    at Function.concat ([stdin]:472:13)
    at Object.<anonymous> ([stdin]:1917:20)
    at Object.6../fileA ([stdin]:1919:4)
    at s ([stdin]:1:254)
    at e ([stdin]:1:425)
    at [stdin]:1:443
    at ContextifyScript.Script.runInThisContext (vm.js:25:33)
    at Object.exports.runInThisContext (vm.js:77:17)
    at Object.<anonymous> ([stdin]-wrapper:6:22)
    at Module._compile (module.js:556:32)
```

feross:
@legokichi What version of browserify are you using? I tested with browserify v14 and could not reproduce the issue. Closing until enough info to reproduce is provided.

feross:
Nevermind -- reinstalling browserify (which pulls in the latest `buffer`) and I can repro this now.

feross:
It turns out that it's not possible to use `instanceof Buffer` safely, like I thought.

It's not possible to use `instanceof Buffer` reliably in a browserify context because there could be multiple different copies of the 'buffer' package in use.

This previous method (checking `buf._isBuffer`) works even for Buffer instances that were created from another copy of the `buffer` package.

NOTE: It's possible to have two different "instances" of the 'buffer' package, even if the 'buffer' package appears only once in the bundled code. This can happen if you require 'buffer' in different ways, for example:

`require('buffer')` vs. `require('buffer/')` vs. using the implicit `Buffer` global.

You can confirm this by browserifying this code:

```js
console.log(require('buffer').Buffer === require('buffer/').Buffer) // will be false
```

[...]

PR #155, opened 2017-02-09
Undo `instanceof Buffer` changes

feross:
It turns out that it's not possible to use `instanceof Buffer` safely, like I thought.

It's not possible to use `instanceof Buffer` reliably in a browserify context because there could be multiple different copies of the 'buffer' package in use.

This previous method (checking `buf._isBuffer`) works even for Buffer instances that were created from another copy of the `buffer` package.

NOTE: It's possible to have two different "instances" of the 'buffer' package, even if the 'buffer' package appears only once in the bundled code. This can happen if you require 'buffer' in different ways, for example:

`require('buffer')` vs. `require('buffer/')` vs. using the implicit `Buffer` global.

You can confirm this by browserifying this code:

```js
console.log(require('buffer').Buffer === require('buffer/').Buffer) // will be false
```

So, for this reason, `instanceof` won't work.

Fixes: https://github.com/feross/buffer/issues/154

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,7 +239,7 @@ function fromArrayBuffer (array, byteOffset, length) {
 }
 
 function fromObject (obj) {
-  if (obj instanceof Buffer) {
+  if (Buffer.isBuffer(obj)) {
     var len = checked(obj.length) | 0
     var buf = createBuffer(len)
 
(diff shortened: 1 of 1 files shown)


feross:
Released as 5.0.5.

index.js
```diff
@@ -474,8 +474,12 @@ function slowToString (encoding, start, end) {
...
 }
 
-// The property is used by the `is-buffer` npm package to detect Buffer instances
-// in Safari 5-7. Remove this eventually.
+// This property is used by `Buffer.isBuffer` (and the `is-buffer` npm package)
+// to detect a Buffer instance. It's not possible to use `instanceof Buffer`
+// reliably in a browserify context because there could be multiple different
+// copies of the 'buffer' package in use. This method works even for Buffer
+// instances that were created from another copy of the `buffer` package.
+// See: https://github.com/feross/buffer/issues/154
```
dcousens: Perfect

dcousens:
LGTM,  nice work @feross.
I did think of this as being an issue when I saw the `instanceof`,  but didn't think anyone would realistically run into the issue since very few people explicitly `require('buffer')`.

feross:
Live and learn!

PR #159, opened 2017-04-05
support environments that lack ArrayBuffer.isView; remove old shims

feross:


feross:
5.0.6

feross:
Added back the `Number.isNaN` shim for IE

PR #165, opened 2017-07-01
Add slice for converting to ArrayBuffer

srsudar:
The .buffer property is an ArrayBuffer, but it is not necessarily the
same length as the original buffer. This adds further documentation to
the README to flesh out converting a Buffer to an ArrayBuffer.

It is based on [your answer here](https://stackoverflow.com/questions/8609289/convert-a-binary-nodejs-buffer-to-javascript-arraybuffer).


feross:
Excellent, thanks!

Issue #167, opened 2017-07-31
iPhone iOS < 10 missing .reverse


rubensayshi:
running on iPhone6 iOS8 and iOS9 simulator fails with reverse being undefined, iOS10 is fine,  
is the lack of iOS < 10 support intentional or could we add a `reverse` polyfill?

(looking at stats for our app we see 2.4% iOS 9.3 and rest is all 10+)

feross:
Sorry -- we can't polyfill this. This is part of the `Uint8Array` API, not part of Buffer. If we started polyfilling every API on `Uint8Array`, this library would get huge. That's out-of-scope for this package as far as I'm concerned. But feel free to include the polyfill separately in your own code.

dcousens:
@feross `Buffer` has `reverse` in Node?

dcousens:
This has caught me out too.

rubensayshi:
Node had `Buffer.reverse`, but it's not documented here: https://nodejs.org/api/buffer.html

dcousens:
@rubensayshi it appears node inherits from `UInt8Array`,  https://github.com/nodejs/node/blob/master/lib/buffer.js#L231,  so as @feross said:

> This is part of the Uint8Array API, not part of Buffer

Frustratingly,  this doesn't really leave us in a fortunate position... I suspect `buffer-reverse` is the only reasonable option to roll-out to ... every repo?

rubensayshi:
"The goal is to provide an API that is 100% identical to node's Buffer API."

To me that sounds like this lib is to ensure any buffer methods available in NodeJS would also be available in browsers..?

PR #168, opened 2017-08-04
Polyfill for Buffer reverse with tests

afk11:
explicitly lists iphone 9 in tests, since it doesn't have Uint8Array.reverse


feross:
Sorry, see comment here: https://github.com/feross/buffer/issues/167#issuecomment-320363831

Issue #166, opened 2017-07-23
Safari: ArrayBuffers are sometimes not `instanceof ArrayBuffer`


jvilk:
## Summary

I'm storing and retrieving ArrayBuffer objects from IndexedDB, and I noticed that in my test environment, the ArrayBuffer objects are not always `instanceof ArrayBuffer` in Safari. As a result, the Buffer package throws the following exception:

> TypeError: First argument must be a string, Buffer, ArrayBuffer, Array,or array-like object.

This behavior does not occur in any other web browser.

## Detailed Explanation

The test environment is that of the Karma test runner, which uses an `iframe` to isolate test code from the runner's code. As a result, the `ArrayBuffer` constructor object in the code's context is not referentially equivalent to the `ArrayBuffer` constructor object in the main window context. This is similar to the age-old `Array` problem, where arrays from other contexts are not necessarily `instanceof Array`, necessitating the use of `Array.isArray`.

Safari chooses the main window's constructor for the `ArrayBuffer` object retrieved from IndexedDB, so it is not an `instanceof ArrayBuffer` in the test code's context.

Here's a screenshot, just so you know I'm not totally mad here. This is evaluated at a breakpoint in the stack frame that throws the `TypeError`:
<img width="394" alt="screen shot 2017-07-22 at 11 06 09 pm" src="https://user-images.githubusercontent.com/1141042/28497093-65cb702e-6f32-11e7-8487-298884496bf1.png">

I do not have a good solution to the problem without resorting to some hack (e.g., is the `byteLength` property present?).

feross:
Interesting... `instanceof` is really quite terrible, isn't it?

Please see this PR with a potential solution: https://github.com/feross/buffer/pull/170 Let me know if that works for you. If so, I'll merge it and do a new release.

jvilk:
> Interesting... instanceof is really quite terrible, isn't it?

I want to agree with you, but I can't really blame JavaScript on this one! `instanceof` is only intended to check if the object was constructed by a particular function, which is totally reasonable. I'd blame the unnecessarily complex native interface between JavaScript code and the web browser, which makes it possible to have multiple constructors for the same type of object! Really, browsers should have a special type of handoff between contexts that appropriately adjusts native objects like this one to use the appropriate constructor object. (Sharing constructors is a no-go for security reasons.)

[...]

PR #170, opened 2017-08-04
Treat ArrayBuffers from another context as valid

feross:
Fixes: https://github.com/feross/buffer/issues/166

ArrayBuffers from another context (i.e. an iframe) do not pass the
`instanceof` check but they should be treated as valid.

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,7 +104,7 @@ function from (value, encodingOrOffset, length) {
     throw new TypeError('"value" argument must not be a number')
   }
 
-  if (value instanceof ArrayBuffer) {
+  if (isArrayBuffer(value)) {
     return fromArrayBuffer(value, encodingOrOffset, length)
   }
 
@@ -364,7 +364,7 @@ function byteLength (string, encoding) {
   if (Buffer.isBuffer(string)) {
     return string.length
   }
-  if (isArrayBufferView(string) || string instanceof ArrayBuffer) {
+  if (isArrayBufferView(string) || isArrayBuffer(string)) {
     return string.byteLength
   }
   if (typeof string !== 'string') {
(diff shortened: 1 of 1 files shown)


feross:
cc @jvilk

jvilk:
@feross do you want to add a test for this case? It looks like you run browser tests. If you can access the DOM during these tests (& only run the test in the browser environment), we can inject an iframe and construct an arraybuffer using its constructor.

jvilk:
Here's an (mostly untested) example demonstrating how this might be accomplished. Due to cross-origin issues, you need to create a data URI iframe that creates another data URI iframe and returns the result to the test context via a `postMessage`.

```javascript
const childFrameHTML = `<head>
<!-- Needs to be the proper URL to buffer. -->
<script type="text/javascript" src="buffer.js"></script>
</head>
<body>
<script type="text/javascript">
  const iframe = document.createElement("iframe");
  iframe.src = 'data:text/html;base64,';
  document.body.appendChild(iframe);
  const ab = new iframe.contentWindow.ArrayBuffer(10);
  const u8arr = new Uint8Array(ab);
  const b = Buffer.from(ab);
  u8arr[0] = 3;
  window.top.postMessage('' + (b[0] === 3), "*");
</script></body>`;
const iframe = document.createElement("iframe");
iframe.onmessage = function(e) {
  console.assert(e.data === "true");
};
iframe.src = 'data:text/html;base64,' + Buffer.from(childFrameHTML, 'utf8').toString('base64');
document.body.appendChild(iframe);
// In cleanup, make sure you document.body.removeChild(iframe);
```

feross:
I think testing for this would probably be pretty brittle. I'm satisfied with just pushing out the fix and relying on code review to ensure that `isArrayBuffer` is always used instead of `instanceof ArrayBuffer`. But if you want to send a PR that works, I would merge it :)

feross:
Released as 5.0.7

Issue #177, opened 2018-01-09
Tests are out of sync with Node.js tests


vmx:
When running `./bin/download-node-tests.js` it's apparent that the tests are heavily out of sync with the Node tests.

I plan to go over them an resync things. I open this issue to prevent double work, so a "go for it" from the maintainers would be appreciated.

ralphtheninja:
I'm not a maintainer, but I'll say "go for it" anyway :grinning:

feross:
Merged @vmx's PR and the tests are closer to node's now, but there's still a lot of differences.

feross:
I just spent like 3 hours merging in more changes from Node.js, but there's still more changes to be done... wow. Would appreciate help from other folks, if there's interest. @vmx, any interest?

I've started adding the new test files that were added. Once they're all added in, then we can delete `test/node/test-buffer.js` which has been removed from the original repo. But I don't want to remove it before the other files are merged in, otherwise we'll lose coverage in the meantime.

vmx:
Thanks for the review @feross. I still have interest and would spend some time next week. I'll start from the top (alphabetically) and work on the differences. I'll push again in small increments to prevent double efforts.

feross:
Great, looking forward to a PR next week, @vmx!

feross:
Note to self: When all these changes are merged in, **publish a new major version.**

superguineapig:
Hey all. It's been some time since this issue was opened and buffer was updated. Has the herd moved to another implementation/fork? Or is there still a roadmap for this one to be actively developed? Just curious.

superguineapig:
Thanks for the info @vmx Glad to hear the project is still alive.

feross:
I'm in grad school as of the past year, so I was off GitHub for a while. Still around and still plan to maintain this as good as ever.

PR #178, opened 2018-01-21
Sync upstream test-buffer-alloc

vmx:
This pull request updates the `test-buffer-alloc.js` from Node.js to the current upstream version.

I've used very granular commits which hopefully shows my workflow. I hope my approach is sound and the work was worth it.

This is part of #177.

### DIFF
--- a/index.js
+++ b/index.js
@@ -188,7 +188,7 @@ function fromString (string, encoding) {
   }
 
   if (!Buffer.isEncoding(encoding)) {
-    throw new TypeError('"encoding" must be a valid string encoding')
+    throw new TypeError('Unknown encoding: ' + encoding)
   }
 
   var length = byteLength(string, encoding) | 0
--- a/index.js
+++ b/index.js
@@ -1462,7 +1462,7 @@ Buffer.prototype.copy = function copy (target, targetStart, start, end) {
   if (targetStart < 0) {
     throw new RangeError('targetStart out of bounds')
   }
-  if (start < 0 || start >= this.length) throw new RangeError('sourceStart out of bounds')
+  if (start < 0 || start >= this.length) throw new RangeError('Index out of range')
   if (end < 0) throw new RangeError('sourceEnd out of bounds')
 
   // Are we oob?
--- a/index.js
+++ b/index.js
@@ -53,6 +53,16 @@ function typedArraySupport () {
   }
 }
 
+Object.defineProperty(Buffer.prototype, 'offset', {
+  enumerable: true,
+  get () {
+    if (!(this instanceof Buffer)) {
+      return undefined
+    }
+    return this.byteOffset
+  }
+})
+
 function createBuffer (length) {
   if (length > K_MAX_LENGTH) {
     throw new RangeError('Invalid typed array length')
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -53,6 +53,16 @@ function typedArraySupport () {
   }
 }
 
+Object.defineProperty(Buffer.prototype, 'offset', {
+  enumerable: true,
```
feross: This should probably be `enumerable: false`, since we don't want these properties to show up in `Object.keys()`. Here's what Node.js does:

```js
Object.keys(Buffer.from([1, 2,3]))
// [ '0', '1', '2' ]
```

index.js
```diff
@@ -53,6 +53,16 @@ function typedArraySupport () {
   }
 }
 
+Object.defineProperty(Buffer.prototype, 'parent', {
+  enumerable: true,
```
feross: Let's remove this so the default value of `enumerable: false` is used.

index.js
```diff
@@ -1560,9 +1560,11 @@ Buffer.prototype.fill = function fill (val, start, end, encoding) {
 // HELPER FUNCTIONS
 // ================
 
-var INVALID_BASE64_RE = /[^+/0-9A-Za-z-_]/g
+var INVALID_BASE64_RE = /[^+/0-9A-Za-z-_=]/g
```
feross: I don't think this line needs to be changed. In the new line you added below, you're already finding the first `=` and discarding it and everything that follows. There should be no `=` left in the string to remove.

index.js
```diff
@@ -1445,6 +1445,9 @@ Buffer.prototype.writeDoubleBE = function writeDoubleBE (value, offset, noAssert
 
 // copy(targetBuffer, targetStart=0, sourceStart=0, sourceEnd=buffer.length)
 Buffer.prototype.copy = function copy (target, targetStart, start, end) {
+  if (target === undefined) {
```
feross: Better to check if the argument is not actually a Buffer to catch all possible bad values instead of just `undefined`

index.js
```diff
@@ -53,6 +53,26 @@ function typedArraySupport () {
...
+    if (!(this instanceof Buffer)) {
+      return undefined
+    }
+    return this.buffer
+  }
+})
+
+Object.defineProperty(Buffer.prototype, 'offset', {
+  enumerable: true,
+  get () {
```
feross: This type of function declaration breaks IE11, fortunately we can just use `get: function () {`

test/common.js
```diff
@@ -0,0 +1,138 @@
...
+'use strict';
+const assert = require('assert');
+const mustCallChecks = [];
+
+function runCallChecks(exitCode) {
+  if (exitCode !== 0) return;
+
+  const failed = mustCallChecks.filter(function(context) {
+    if ('minimum' in context) {
+      context.messageSegment = `at least ${context.minimum}`;
```
feross: The template literal syntax breaks IE11.

test/common.js
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  if (failed.length) process.exit(1);
+}
+
+exports.mustCall = function(fn, exact) {
+  return _mustCallInner(fn, exact, 'exact');
+};
+
+function _mustCallInner(fn, criteria = 1, field) {
```
feross: Default argument values break IE11

feross:
I figured out the reason that IE11 was unhappy -- `test/common.js` belongs in `test/node/common.js`. The files that match `test/*` run in both ES5 and ES6+ browsers, while `test/node/*` only run in ES6+ browsers. Moving the file into `node/` fixes the IE11 errors.

feross:
Fixed the other issues I mentioned and merged this. Thanks for doing this work!

Issue #174, opened 2017-09-18
Buffer.prototype.copy


blake-regalia:
Forgive me if this was already addressed in another thread but it strikes me as odd.

It seems we should take advantage of [TypedArray#copyWithin](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/copyWithin) for the `this === target` conditional branch in the current [Buffer.prototype.copy](https://github.com/feross/buffer/blob/a50b8e35991255b21dbad23a189fd32d66b6868c/index.js#L1476). Why is it using a for loop instead?

Also, why is the next branch testing for `len < 1000` ? Was there a benchmark done to arrive at this number? It seems arbitrary... I believe TypedArray#set is appropriate for all cases that are not copyWIthin.

I can PR this if we want.

feross:
Good suggestion.

This package was written before `copyWithin` existed. We can use it optimistically, but we still need a fallback for IE11 and Safari 8. See [MDN compat table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/copyWithin#Browser_compatibility).

I ran a benchmark and arrived at `1000` below which set performance was worse, but honestly, browser engines have changed a lot in the last few years, so I'm fine with removing that special case and just leaving it to browsers to worry about the performance of `set`.

feross:
Okay, I went ahead and made these changes!

PR #183, opened 2018-02-16
Use copyWithin when available

blake-regalia:
#174


feross:
Ahh, sorry I should have mentioned in the issue comments that I was going to do this. I already committed my solution to this. Apologies! Thanks for the issue and PR!! ❤️

PR #194, opened 2018-04-27
Add TypeScript declaration file

ukstv:


feross:
I would publish this immediately, but we're blocked on https://github.com/feross/buffer/issues/177 since we actually have some breaking changes merged on `master`. Once those are all merged in, I'll released this as a new major version.

brunolemos:
@feross you could checkout to the last published commit and add only these types there; came here looking for types.

PR #204, opened 2018-06-20
Use `util.inspect.custom` symbol if available.

goto-bus-stop:
Ports the `buffer-inspect` test from Node core.
Adds a `Buffer.prototype[util.inspect.custom]` method, which is an alias
to `Buffer.prototype.inspect`. Node already doesn't have an `.inspect`
method on Buffers anymore, but since this module has to work in browsers
that do not have Symbols, it seems better to keep it around.

In Node, this will use the builtin `util.inspect.custom` symbol. In the
browser, it will use `Symbol.for('util.inspect.custom')`. The browser
version of `util` will also use the `inspect-custom-symbol` module in
the near future.

If https://github.com/nodejs/node/pull/20857 gets merged,
`Symbol.for('util.inspect.custom')` will be used everywhere and the
dependency on `inspect-custom-symbol` could probably be dropped.

The motivation for this is API parity and the fact that Node is
removing support for the old `.inspect` method:
https://github.com/nodejs/node/pull/20722


feross:
Hi @goto-bus-stop,

I'm down to merge this PR. But I noticed that Node.js ended up going with `Symbol.for('nodejs.util.inspect.custom')` and the https://github.com/mafintosh/inspect-custom-symbol package still uses `Symbol.for('util.inspect.custom')`. What do you think about just skipping the package using `Symbol.for` directly?

Thanks for pointing out that `buffer.inspect` has been deprecated. We can remove that in another PR. I opened an issue here: https://github.com/feross/buffer/issues/239

goto-bus-stop:
Will try to update later today :+1:

feross:
5.4.0

fanatid:
Not sure how this relevant and correct, but `Symbol.for` was available from Chrome>=40, while `Symbol` was from Chrome>=38
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Browser_compatibility

Maybe?
```js
typeof Symbol !== 'undefined' && typeof Symbol.for === 'function'
```

feross:
@fanatid Nice catch. However, Chrome 40 was released on 2015-01-21. I don't think anyone is still running that version, and we only officially support the last 2 versions of Chrome.

ljharb:
`Symbol.for` is a syntax error in older browsers; I filed #254 to use bracket notation to avoid that.

PR #209, opened 2018-07-14
fix/cast boolean to unsigned with alloc

crabicode:
Cast booleans to unsigned when `Buffer.alloc(1, true)`, which is a default behaviour in nodejs v8.

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,8 +198,9 @@ function alloc (size, fill, encoding) {
     // prevents accidentally sending in a number that would
     // be interpretted as a start offset.
     return typeof encoding === 'string'
-      ? createBuffer(size).fill(fill, encoding)
-      : createBuffer(size).fill(fill)
+      // cast boolean to unsigned
+      ? createBuffer(size).fill(typeof fill === 'boolean' ? Number(fill) : fill, encoding)
+      : createBuffer(size).fill(typeof fill === 'boolean' ? Number(fill) : fill)
   }
   return createBuffer(size)
 }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -299,6 +300,10 @@ function fromObject (obj) {
     return buf
   }
 
+  // if (typeof obj === 'boolean') {
```
crabicode: already removed this comment

test/node/test-buffer-alloc.js
```diff
@@ -182,6 +182,17 @@ assert.doesNotThrow(() => Buffer.alloc(1).write('', 1, 0));
...
+{
+  // cast boolean to unsigned
+  const a = Buffer.alloc(1, true);
+  assert.ok(a.length === 1);
+  assert.ok(a[0] === 1);
+
+  const b = Buffer.alloc(1, false);
+  assert.ok(b.length === 1);
+  assert.ok(b[0] === 0);
+}
```
feross: Does this test case come from Node.js core?

index.js
```diff
@@ -198,8 +198,9 @@ function alloc (size, fill, encoding) {
     // prevents accidentally sending in a number that would
     // be interpretted as a start offset.
     return typeof encoding === 'string'
-      ? createBuffer(size).fill(fill, encoding)
-      : createBuffer(size).fill(fill)
+      // cast boolean to unsigned
+      ? createBuffer(size).fill(typeof fill === 'boolean' ? Number(fill) : fill, encoding)
+      : createBuffer(size).fill(typeof fill === 'boolean' ? Number(fill) : fill)
```
feross: Instead of repeating the `typeof fill === 'boolean'` twice, can you pull this out into a line above the return statement?

```js
if (typeof fill === 'boolean') fill = Number(fill)
```

PR #221, opened 2018-11-27
Update LICENSE

r-browser-app-team:
MIT license needs both of year and copyright holder.
Please complete MIT license. We want to use this without legal risk.


feross:
Thanks for the PR, but I disagree.

Issue #244, opened 2019-09-10
readBigUInt64BE() and similar methods are missing


nazar-pc:
Essentially getting this in Browserify build:
> TypeError: Buffer.from(...).readBigUInt64BE is not a function

feross:
According to Node docs this was added in Node 12.0.0 so it makes sense that it's missing from here.

nazar-pc:
Yes, but v12 is out for a while now, so I think it is time to consider implementing new methods.

feross:
I'd love to do it, but I'm very time-constrained at the moment. PR welcome, from you or anyone else!

fanatid:
Node.js internally uses `BigInt` constructor, if this package will start use it... should this methods be available if BigInt is not available in environment?
I'm not familiar too much with bundling systems, can usage of BigInt cause any issues with it?

fanatid:
Just checked how this implemented in node. I think better finish #217 first. @feross are you ok with this?
For reference, PR in node: https://github.com/nodejs/node/pull/19691

feross:
Yep, feel free to tackle this in any order your prefer. Finishing #217 first sounds good. Thanks for all your help, @fanatid.

feross:
Released as 6.0.0

PR #247, opened 2019-09-19
[WIP] add rw BigInt64 methods

fanatid:
Original issue #244 
This is quick copy code from node. I did not want make everything perfect because have a questions.

Is it ok use let/const/arrow functions/extends/template strings/etc? What if:
1) Package will start use latest syntax and will be transpiled to what should be supported?
2) Package will start use latest syntax and distribute it, so anybody who use it will need traspile to what they actually use? This is not used usually, right? But from my understanding this should give best code for projects who uses buffer, because they will receive only transpiled parts which they not support in their environment.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1187,6 +1187,48 @@ Buffer.prototype.readUInt32BE = function readUInt32BE (offset, noAssert) {
     this[offset + 3])
 }
 
+Buffer.prototype.readBigUInt64LE = defineBigIntMethod(function readBigUInt64LE (offset = 0) {
+  validateNumber(offset, 'offset')
+  const first = this[offset]
+  const last = this[offset + 7]
+  if (first === undefined || last === undefined) {
+    boundsError(offset, this.length - 8)
+  }
+
+  const lo = first +
+    this[++offset] * 2 ** 8 +
+    this[++offset] * 2 ** 16 +
+    this[++offset] * 2 ** 24
+
+  const hi = this[++offset] +
+    this[++offset] * 2 ** 8 +
+    this[++offset] * 2 ** 16 +
+    last * 2 ** 24
+
+  return BigInt(lo) + (BigInt(hi) << 32n)
+})
+
+Buffer.prototype.readBigUInt64BE = defineBigIntMethod(function readBigUInt64BE (offset = 0) {
+  validateNumber(offset, 'offset')
+  const first = this[offset]
+  const last = this[offset + 7]
+  if (first === undefined || last === undefined) {
+    boundsError(offset, this.length - 8)
+  }
+
+  const hi = first * 2 ** 24 +
+    this[++offset] * 2 ** 16 +
+    this[++offset] * 2 ** 8 +
+    this[++offset]
+
+  const lo = this[++offset] * 2 ** 24 +
+    this[++offset] * 2 ** 16 +
+    this[++offset] * 2 ** 8 +
+    last
+
+  return (BigInt(hi) << 32n) + BigInt(lo)
+})
+
 Buffer.prototype.readIntLE = function readIntLE (offset, byteLength, noAssert) {
   offset = offset >>> 0
   byteLength = byteLength >>> 0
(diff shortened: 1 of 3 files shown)


feross:
> Is it ok use let/const/arrow functions/extends/template strings/etc?

If we want to continue supporting IE11 that means we can use:

- [let](https://caniuse.com/#feat=let)
- [const](https://caniuse.com/#feat=const) in most places

But should probably avoid the rest.

If we want to use new syntax in the codebase, that would probably let us port code from Node.js more easily, which I like.

But then we need to either:

- drop IE11 and publish a new major version
- add `babel` and start shipping a transpiled build too

fanatid:
Why if we add `babel` we need drop IE11? Not all polyfills work in IE11? Major bump because too much changes?

feross:
@fanatid I meant that we need to do one of those two option.

If we start using new syntax without babel then this package stops working in IE11 unless users start transpiling it. So that's a breaking change and we need to publish a new version.

Or, we use new syntax, add `babel`, and ship a transpiled build. Then no new major version is needed because nothing breaks from the user's perspective.

vweevers:
It might be harder to optimize a transpiled build for performance?

feross:
Yeah, true.

I'm happy to just optimize for newer browsers. The current version of `buffer` works fine in IE11 and folks can always continue using that version if they need IE11 support. We can continue development on a new major version. Let's do it.

fanatid:
Thanks for explanation Feross.
@vweevers probably yes, but continue manually port new code to old syntax is pain for everybody. Other thing that not traspiled code should be executed faster than manually ported because code should be more effective? (but this will break almost all projects who uses buffer? haha)

So, decision is major version bump with new JS features transpiled to supported browsers? :smile:

feross:
I think we should not transpile. Users can do that if they want. We can release a new major version that includes new language features. We shouldn't get carried away, though. Tests should still pass in the last two major versions of every major browser (Chrome, Firefox, Safari, Edge, Safari on iOS, Chrome on Android).

java-james:
When do you think we can expect resolution and merge of this PR?

fanatid:
Sorry, do not have time right now to finish it. Maybe later.

collincusce:
Heyyy do you have time now? This is a feature I could immediately use! <3

collincusce:
@fanatid or @feross there's a lot of missing functions in Buffer and classes like, BN.js which has toArrayLike and tries to use Buffer fails. If this fix patches that, can we get it reviewed and merged?

fanatid:
Can you describe situations (or show examples) when bn.js fails with Buffer?

collincusce:
Absolutely, I should definitely be more clear. I apologize. Using the current nodejs LTS + Typescript, typechecking fails because these functions are missing. This isn't just about BN.js but also things like elliptic. Using any Buffer as a parameter will give an error. Here's some examples of type conflicts. 


Screenshot from 2019-12-14 13-38-15

Screenshot from 2019-12-14 13-29-16

Screenshot from 2019-12-14 13-28-11

fanatid:
Ah, so it's mostly about TypeScript?

collincusce:
Yes sorry if that's not a roadmap item, but it definitely would fix Typescript issues if this push was put in.

collincusce:
Any updates?

Tschakki:
@feross could you merge this one please? :)

feross:
I'm okay with shipping a new version of `buffer` only for modern browsers. We can consider shipping a transpiled version if we want to preserve current browser support, but let's do that in another issue.

@fanatid Do you think this is ready to merge?

feross:
I just tried running the ES6 browser tests and they throw an exception because the way that `assert.throws` is used here:

Screen Shot 2020-06-17 at 5 54 38 PM

Can someone figure out the cause of this? If you do, please either update this PR (@fanatid) or send a PR to this PR (anyone else).

collincusce:
> I'm okay with shipping a new version of buffer only for modern browsers. 

Please! I'm doing so many hokey things to handle Uint64 types lol, and the real barrier here is just getting TypeScript to agree that it's a Buffer class. Without the functions matching, TypeScript balks. Can we make it match the latest nodejs LTS version?

fanatid:
Sorry, I do not have a time to check/update changes =(. Will be happy if somebody fork PR with required fixes / changes.

shuse2:
@feross @fanatid 
I forked thie PR and add a commit to original commit in https://github.com/feross/buffer/pull/267 to address the issue https://github.com/feross/buffer/pull/247#issuecomment-645646172

Running `npm run test-browser-es6-local`, it does not show the thrown error, and still passes the test.

Please check when you have time

feross:
Thanks @shuse2 – let's use #267

feross:
Released as 6.0.0

PR #249, opened 2019-10-17
Add enterprise language

phated:
Our beta test indicated that having this language near the top was the most successful. @feross  Please let me know if the location I chose seems out of place.


feross:
Sure, let's try it out

PR #250, opened 2019-10-24
Different behaviour from node.js for utf16leSlice

roaringUnicorn:
# Problem
```js
const oddLengthBuffer = Buffer.from('abcde', 'utf8')
console.log(Buffer.from(oddLengthBuffer.toString('utf16le'), 'utf16le').toString('utf8')) // ‘abcd’ in Node.js and ‘abcd\x00\x00’ with polyfill
```
This comes from accessing n+1-th element of a Buffer of size n in utf16leSlice.
# Solution
We omit the last byte for an odd length buffer when encoding to UTF16-LE

### DIFF
--- a/index.js
+++ b/index.js
@@ -1076,7 +1076,8 @@ function hexSlice (buf, start, end) {
 function utf16leSlice (buf, start, end) {
   var bytes = buf.slice(start, end)
   var res = ''
-  for (var i = 0; i < bytes.length; i += 2) {
+  // If bytes.length is odd, the last 8 bits must be ignored (same as node.js)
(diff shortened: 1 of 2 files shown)


gyzerok:
Hello @roaringUnicorn!

I seem to get a similar issue. Can you confirm that this is the source of it?

```js
// Node
Buffer.from([0x61]).toString('utf16le') == ''
// feross/buffer
Buffer.from([0x61]).toString('utf16le') == '\u0000'
```

feross (approved):
LGTM

feross:
Released as 5.7.0.

PR #253, opened 2019-12-07
Add fromArrayView for TypedArrays and DataViews. Fast path for Uint8A…

koush:
I'm using webtorrent in Android (via a quickjs runtime that embeds into java).

I noticed that performance was extremely slow, and found that the torrent piece byte arrays were being copied here, in a JS for loop, treating Uint8Arrays as if they were just normal arrays. Pieces are a couple MB, so every time Buffer.from was called on a piece, it would loop millions of times, in JavaScript, to do the copy.

Given that the underlying Array view is a Uint8Array, it's possible to use the ArrayBuffer fast path here, and allow the runtime to do the memcpy (be it browser or quickjs).

This patch ended up resolving all my perf issues.

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,7 +127,7 @@ function from (value, encodingOrOffset, length) {
   }
 
   if (ArrayBuffer.isView(value)) {
-    return fromArrayLike(value)
+    return fromArrayView(value)
   }
 
   if (value == null) {
@@ -267,6 +267,14 @@ function fromArrayLike (array) {
     buf[i] = array[i] & 255
   }
   return buf
+}
+
+function fromArrayView (arrayView) {
+  if (isInstance(arrayView, Uint8Array)) {
+    var copy = Uint8Array.from(arrayView)
+    return fromArrayBuffer(copy.buffer, copy.byteOffset, copy.byteLength)
+  }
+  return fromArrayLike(arrayView)
 }
 
 function fromArrayBuffer (array, byteOffset, length) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -269,6 +269,12 @@ function fromArrayLike (array) {
   return buf
 }
 
+function fromArrayView (arrayView) {
+  if (isInstance(arrayView, Uint8Array))
```
koush: Int8Array may also be a candidate.

feross (approved):
LGTM

index.js
```diff
@@ -269,6 +269,14 @@ function fromArrayLike (array) {
   return buf
 }
 
+function fromArrayView (arrayView) {
+  if (isInstance(arrayView, Uint8Array)) {
+    var copy = Uint8Array.from(arrayView)
```
feross, line 274: To increase compatibility with browsers that don't have `Uint8Array.from` I changed this line to:

`var copy = new Uint8Array(arrayView)`

I believe this shouldn't affect performance, but it would be great if you could confirm this, @koush.

feross:
Released as 5.7.0.

feross:
@koush It appears that your PR has introduced a bug into `buffer`: https://github.com/feross/buffer/issues/272 Do you think you can send a PR to fix it?

koush:
I'll take a look this week.

feross:
@koush A fix was sent in https://github.com/feross/buffer/pull/273. If you get a chance, please confirm that it looks good to you and doesn't regress the performance improvements you committed.

PR #254, opened 2019-12-20
[Fix] `Symbol.for` is a syntax error in older browsers

ljharb:
For example, Opera 10.6.

Introduced in #204; affects v5.4.0+.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,8 +11,8 @@
 var base64 = require('base64-js')
 var ieee754 = require('ieee754')
 var customInspectSymbol =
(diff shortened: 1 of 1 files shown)


ljharb:
It seems like buffer v5 doesn't work on old browsers in general, so while I think this would be a good fix, I understand if it's not accepted.

feross:
Thanks for the PR – sorry for the delay in reviewing this. Only issue is a lint error. I will fix post-merge.

feross (approved):

ljharb:
Awesome, thanks :-)

feross:
Released as 5.7.0.

Issue #256, opened 2020-01-14
Buffer.from with SharedArrayBuffer?


corwin-of-amber:
According to [the docs](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length), `Buffer.from(arrayBuffer, byteOffset, length)` accepts either an `ArrayBuffer` or a `SharedArrayBuffer` as the first argument.

This implementation's `Buffer.from` only checks if the argument is an `ArrayBuffer`. `SharedArrayBuffer` is *not* a subclass of `ArrayBuffer` (possibly because ArrayBuffers can be detached, while SharedArrayBuffers cannot?). So `Buffer.from(sab)` fails if `sab` is an instance of `SharedArrayBuffer`.

PR #259, opened 2020-02-25
Fix Buffer.from for SharedArrayBuffer

mischnic:
Closes https://github.com/feross/buffer/issues/256

`SharedArrayBuffer` has the same properties as `ArrayBuffer` (`new Uint8Array(sab)` works, `sab.byteLength` exists), therefore `fromArrayBuffer` can be used.

### DIFF
--- a/index.js
+++ b/index.js
@@ -142,6 +142,11 @@ function from (value, encodingOrOffset, length) {
     return fromArrayBuffer(value, encodingOrOffset, length)
   }
 
+  if (isInstance(value, SharedArrayBuffer) ||
+      (value && isInstance(value.buffer, SharedArrayBuffer))) {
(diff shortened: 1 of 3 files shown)


VictorQueiroz:
@feross Do we have an expectation to get this through? Very helpful for those like me, working with SharedArrayBuffer.

mischnic:
@feross ?

feross:
Published as 5.5.0.

mischnic:
Thank you! (But I still don't see the version on npm)

feross:
Note: Before publishing 5.5.0, I had to add back support for browsers which lack `SharedArrayBuffer` support in this commit: https://github.com/feross/buffer/commit/ae2ab6740394bc60c0e28fc5eca67b2e663a688a

It's published now.

mischnic:
👍

PR #262, opened 2020-04-03
dev: ignore package-lock.json

mikeal:
I noticed that you don’t checkin the lock file but also haven’t added it to .gitignore, which is a bit annoying during development.


mikeal:
apologies, didn’t realize this was still on top of the branch with the other fix.

PR #263, opened 2020-04-14
Remove unneeded Symbol.species code

### LINKED ISSUE #260: .subarray method should return `Buffer` instance
Ran across this one today.

`Buffer.subarray()` in Node.js will return a `Buffer` instance, but in the browser it returns a `Uint8Array`.

feross:
We now handle calls like:

new Buffer(arraybuffer, 2, 8) so there's no need for this workaround originally implemented in: https://github.com/feross/buffer/pull/97

More info: https://bugs.chromium.org/p/v8/issues/detail?id=4665#c4

Fixes #260. Supersedes #261

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,17 +108,6 @@ function Buffer (arg, encodingOrOffset, length) {
   return from(arg, encodingOrOffset, length)
 }
 
-// Fix subarray() in ES2016. See: https://github.com/feross/buffer/pull/97
-if (typeof Symbol !== 'undefined' && Symbol.species != null &&
-    Buffer[Symbol.species] === Buffer) {
-  Object.defineProperty(Buffer, Symbol.species, {
-    value: null,
-    configurable: true,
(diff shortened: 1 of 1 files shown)


mikeal:
Test still fails for me.

Here’s a super minimal test you can run yourself.

```js
const { Buffer } = require('buffer')
const assert = require('assert')
it('subarray', done => {
  const test = Buffer.from('asdf')
  const sub = test.subarray(2,3)
  assert.ok(Buffer.isBuffer(sub))
  done()
})
```

This will fail:

```
npx polendina —cleanup test.js
````

mikeal:
You know what, I take it back, this does fix it, the module just isn’t being picked up properly when I use a git remote branch in the package.json. If I change the import to a relative path where I’ve checked out this branch the test passes.

👍

feross:
Released as 5.6.0

PR #265, opened 2020-05-01
Remove descending copy from end

seishun:
Uint8Array.prototype.set handles overlapping ranges on its own.


feross (commented):
LGTM

feross (approved):

feross:
Released as 5.7.0.

PR #267, opened 2020-06-21
Add RW BigInt64 methods

shuse2:
### Descriptions

This PR is extension of PR #247 by @fanatid. 
It addresses the issue described in https://github.com/feross/buffer/pull/247#issuecomment-645646172

### DIFF
--- a/index.js
+++ b/index.js
@@ -1187,6 +1187,48 @@ Buffer.prototype.readUInt32BE = function readUInt32BE (offset, noAssert) {
     this[offset + 3])
 }
 
+Buffer.prototype.readBigUInt64LE = defineBigIntMethod(function readBigUInt64LE (offset = 0) {
+  validateNumber(offset, 'offset')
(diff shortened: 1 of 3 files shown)


feross:
Looks good. Since this PR uses the exponentiation operator, I'm going to drop support for IE11 and Safari 9-10 in order to accept it.

Further changes I needed to make:

- Use consistent shifting to floor the offset numbers, like we do for other methods
- Remove all BigInt literals in favor of using a BigInt constructor – this fixes parse errors in older browsers like Safari 11
- Disable Node tests in Safari 11 and 12 since they use BigInt literal syntax

feross:
Released as 6.0.0

jimmywarting:
> I'm going to drop support for IE11 and Safari 9-10 in order to accept it.

Could this mean that we could do things such as `class Buffer extends Uint8Array { ... }` instead?

PR #269, opened 2020-08-07
Add codespell action

peternewman:


feross:
Thanks for fixing the typos. Don't want to maintain any more things that can break/require maintenance, so I removed the Codespell action.

feross (approved):

PR #273, opened 2020-11-04
add length check in Buffer.concat

jkkang-kosac:


feross (approved):

feross:
Thanks – very helpful PR!

feross:
Released as 5.7.1 and 6.0.1.

PR #277, opened 2020-11-21
Types: Using Uint8Array instead of Buffer

martinheidegger:
The API supports at various places to use Uint8Arrays to be used instead of Buffers. This PR loosens the type restrictions on a few instances where I noticed that an Uint8Array would suffice.


feross (approved):

feross:
Released as 6.0.3

PR #285, opened 2020-12-31
Make Buffer more compliant with official implementation

Soremwar:
This fixes all the instances where Uint8Array is a valid parameter but errors due to the parameter "not being a buffer"

### DIFF
--- a/index.js
+++ b/index.js
@@ -609,7 +609,7 @@ Buffer.prototype.toString = function toString () {
 Buffer.prototype.toLocaleString = Buffer.prototype.toString
 
 Buffer.prototype.equals = function equals (b) {
-  if (!Buffer.isBuffer(b)) throw new TypeError('Argument must be a Buffer')
+  if (!Buffer.isBuffer(b) && !(b instanceof Uint8Array)) throw new TypeError('Argument must be a Buffer')
   if (this === b) return true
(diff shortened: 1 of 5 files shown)


Soremwar:
@feross I know this is a small change, but this is needed to land https://github.com/nodejs/readable-stream/pull/452 so your review will be very appreciated

addaleax:
@feross Any chance you could take a look at this? :)

addaleax:
ping @feross? :)

index.js
```diff
@@ -1688,7 +1688,7 @@ Buffer.prototype.writeDoubleBE = function writeDoubleBE (value, offset, noAssert
 
 // copy(targetBuffer, targetStart=0, sourceStart=0, sourceEnd=buffer.length)
 Buffer.prototype.copy = function copy (target, targetStart, start, end) {
-  if (!Buffer.isBuffer(target)) throw new TypeError('argument should be a Buffer')
+  if (!Buffer.isBuffer(target) && !(target instanceof Uint8Array)) throw new TypeError('argument should be a Buffer')
```
addaleax, line 1691: It’s probably nicer to use `Object.prototype.toString.call(target) === '[object Uint8Array]'` here instead of `instanceof`, because `instanceof` still won’t work across multiple realms (e.g. iframes)

PR #298, opened 2021-06-18
Redefined subarray in index.d.ts to return Buffer

### LINKED ISSUE #297: subarray is not re-defined in index.d.ts
The `subarray` function should return a `Buffer`, but currently the types are inheriting the definition from `Uint8Array` which has the function returning a `Uint8Array`.
I believe PR #261 fixed this issue at runtime, but the types were never updated.

BTOdell:
Closes #297


kinetikeith:
Could this pull be reviewed? It's been open since 2021.

kinetikeith (approved):
Declaration matches with [official docs](https://nodejs.org/api/buffer.html#bufsubarraystart-end) as well as the local implementation

dcousens (approved):

PR #303, opened 2021-10-16
Fix fromHex invalid hex behavior

junderw:
In NodeJS, the behavior of Buffer.from(x, 'hex') differed slightly on its handling of bad hex parsing.

My fix also increased performance. I tried parsing a hex string `'7c'.repeat(5e7)` and it went from 2.8 s to ~800ms.

```node
// node
> Buffer.from('abc def01','hex')
<Buffer ab>
```

```node
// this library
> require('./index.js').Buffer.from('abc def01','hex')
<Buffer ab 0c de f0>
```


junderw:
Note: This new implementation also mimics the code for NodeJS hex parser used in the `from(x, 'hex')` portion.

https://github.com/nodejs/node/blob/v14.18.1/src/string_bytes.cc#L246-L261

index.js
```diff
@@ -844,9 +844,12 @@ function hexWrite (buf, string, offset, length) {
   }
   let i
   for (i = 0; i < length; ++i) {
-    const parsed = parseInt(string.substr(i * 2, 2), 16)
-    if (numberIsNaN(parsed)) return i
-    buf[offset + i] = parsed
+    const a = hexCharValueTable[string[i * 2]]
```
Uzlopak, line 847: why not bitshift here?

junderw: Because it is in line with the style of the code it replaces.

Also, there is no performance difference.

dcousens (approved):

test/node/test-buffer-badhex.js
```diff
@@ -14,6 +14,12 @@ const assert = require('assert');
   assert.strictEqual(buf.write('abcdef01', 0, 'hex'), 4);
   assert.deepStrictEqual(buf, new Buffer([0xab, 0xcd, 0xef, 0x01]));
   assert.strictEqual(buf.toString('hex'), 'abcdef01');
+  // Node Buffer behavior check
+  // > Buffer.from('abc def01','hex')
+  // <Buffer ab>
+  assert.strictEqual(buf.write('abc def01', 0, 'hex'), 1);
+  assert.deepStrictEqual(buf, new Buffer([0xab]));
+  assert.strictEqual(buf.toString('hex'), 'ab');
```
LinusU, line 22: `buf` here is a 4 bytes long buffer, so these asserts can never succeed? Did you try this locally? 🤔

LinusU:
This broke both the linting and the tests, I've submitted PRs to fix that: #340, #341 

ping @dcousens

test/node/test-buffer-badhex.js
junderw: buf should NOT be a 4 byte buffer. That's the point.

If it is a 4 byte buffer, than the code is broken.

a non-hex character should end the hex parsing and only the 2-hex pairs up until the first non-hex is output as a Buffer.

in this case, `[0xab]`

LinusU: But you are doing `.write` on an already existing `Buffer` that is 4 bytes long. `write` cannot truncate the length of the buffer.

With my updated tests it checks that `write` only wrote 1 byte, and that it left the three bytes after as `00 00 00`.

dcousens:
With CI passing now, I'm slightly more confident with this going out (see https://github.com/feross/buffer/issues/335)

chjj:
@dcousens, please look at #343 before tagging a release. This PR (#303) added a massive slowdown.

PR #312, opened 2022-04-26
add missing constants

### LINKED ISSUE #311: Missing buffer.constants.*
There appears to be a missing implementation of 
`buffer.constants.MAX_LENGTH` & `buffer.constants.MAX_STRING_LENGTH` that is documented in the [node buffer api](https://nodejs.org/docs/latest-v14.x/api/buffer.html#buffer_buffer_constants)

urg:
add missing kStringMaxLength, constants.MAX_LENGTH & constants.MAX_STRING_LENGTH

fixes #311

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,16 @@ exports.INSPECT_MAX_BYTES = 50
 const K_MAX_LENGTH = 0x7fffffff
 exports.kMaxLength = K_MAX_LENGTH
 
+// not used, but value is added for maintain api compatability
+// max length will vary from browser to browser, but using a likely expected value circa Node v8
+const K_STRING_MAX_LENGTH = 2**28 - 1
+exports.kStringMaxLength = K_STRING_MAX_LENGTH
+
+exports.constants = {
+  MAX_LENGTH: K_MAX_LENGTH,
+  MAX_STRING_LENGTH: K_STRING_MAX_LENGTH
+}
+
 /**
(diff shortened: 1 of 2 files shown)


dcousens (changes requested):
LGTM, but I want a reference for that `2**28 - 1` value.
Reading https://nodejs.org/docs/latest/api/buffer.html#bufferconstantsmax_length , the value seems to be about `2**30-1` on 32-bit architectures.

I think any `buffer` actually near these magnitudes will probably fail no matter, but, if we have a value, it should have a reference.

urg:
@dcousens  - good point on the reference. I can't find where I got the initial value from, however, its now set to what it is in the current version of node.

Please note this is for the max string length and not the max buffer length (https://nodejs.org/docs/latest/api/buffer.html#bufferconstantsmax_string_length). Max buffer length was already defined and just exported to match api spec.

dcousens (approved):

dcousens:
Thanks for the fast follow up @urg, LGTM!

dcousens:
The tests are passing, the CI has other issues (see https://github.com/feross/buffer/pull/332, I'll be resolving this before the next release).

jeremymeng:
Looks that the typing file hasn't been upgraded to include the newly added constants.

dcousens:
@jeremymeng I have this on my TODO, but I don't have an ETA for you

PR #324, opened 2023-04-15
Accept Uint8array for copy target

tyfkda:
It is convenient if `copy` method accept `Uint8Array` as a target buffer.

I'm using `memfs` module
and I wonder if `read` method can accept `Uint8Array` directly:
https://github.com/streamich/memfs/blob/8d9e5332efb800a673f5f3c9702817b7d359ce1a/src/node.ts#L135


tyfkda:
I confirmed the test code works on native `Buffer` provided by node.js, too (v18.14.0).

```js
const buf1 = new Buffer(26)
for (let i = 0; i < 26; i++) {
  buf1[i] = i + 97 // 97 is ASCII a
}

const u8buf = new Uint8Array(26)
u8buf.fill(35) // ASCII #

buf1.copy(u8buf, 8, 20, 24)
console.log(new TextDecoder().decode(u8buf.slice(0, 25)))
```

PR #327, opened 2023-05-30
Update README.md

awerlang:
Update notice about node.js compatibility


dcousens:
The difference was removed in https://github.com/feross/buffer/pull/143, thanks @awerlang

dcousens (approved):

PR #340, opened 2024-01-23
Fix linting errors

LinusU:
Fixes some Standard linting errors that was introduced in #303


junderw:
This lint failed because it was added to `standard` after I made the PR years ago, and standard is using `"*"` as the version in package.json

LinusU:
(failing CI adressed in #342)

dcousens (approved):

PR #341, opened 2024-01-23
Fix failing tests for #303 (unreleased)

LinusU:
Fixes the failing tests introduced in #303. Since the `Buffer` was 4 bytes long these comparisons couldn't ever work before.

### DIFF
--- a/test/node/test-buffer-badhex.js
+++ b/test/node/test-buffer-badhex.js
@@ -21,6 +21,7 @@ const assert = require('assert');
   assert.strictEqual(buf.write('abc def01', 0, 'hex'), 1);
   assert.deepStrictEqual(buf, new Buffer([0xab, 0, 0, 0]));
   assert.strictEqual(buf.toString('hex'), 'ab000000');
+  assert.deepStrictEqual(Buffer.from('abc def01', 'hex'), Buffer.from([0xab]));
 
   const copy = Buffer.from(buf.toString('hex'), 'hex');
   assert.strictEqual(buf.toString('hex'), copy.toString('hex'));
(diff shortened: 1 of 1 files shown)


junderw:
This is incorrect.

We expect the buffer to be 1 byte.

test/node/test-buffer-badhex.js
```diff
@@ -17,9 +17,10 @@ const assert = require('assert');
   // Node Buffer behavior check
   // > Buffer.from('abc def01','hex')
   // <Buffer ab>
+  assert.strictEqual(buf.write('00000000', 0, 'hex'), 4);
   assert.strictEqual(buf.write('abc def01', 0, 'hex'), 1);
-  assert.deepStrictEqual(buf, new Buffer([0xab]));
-  assert.strictEqual(buf.toString('hex'), 'ab');
+  assert.deepStrictEqual(buf, new Buffer([0xab, 0, 0, 0]));
+  assert.strictEqual(buf.toString('hex'), 'ab000000');
```
junderw, line 23: Just replace these 4 lines with the following.

```js
  assert.deepStrictEqual(Buffer.from('abc def01', 'hex'), Buffer.from([0xab]));
```

LinusU:
> This is incorrect.
> 
> We expect the buffer to be 1 byte.

This is doing `.write` to a `Buffer` that is already 4 bytes long. `write` cannot truncate the buffer so the buffer will still be 4 bytes...

test/node/test-buffer-badhex.js
LinusU: Probably best to keep both? To make sure that the `write` function doesn't write more than 1 byte?

LinusU: I've pushed a commit that adds your suggested line as well

LinusU:
(failing CI adressed in #342)

dcousens:
@LinusU can you rebase this?

LinusU:
@dcousens done ✅

PR #342, opened 2024-01-23
Remove failing and redundant Standard workflow

LinusU:
There is no script called `standard` so `npm run standard` was always failing.

Also, `standard` is already run as part of `npm test`, so it's covered by the other workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,15 +22,3 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - run: npm install
     - run: npm test
-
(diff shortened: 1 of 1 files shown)


dcousens (approved):

dcousens:
Thanks @LinusU , but I think we should merge https://github.com/feross/buffer/pull/332 instead

dcousens:
Added https://github.com/feross/buffer/commit/3404b7e56bd6841b50afd3bc5664d5f4c8d77c2a to resolve the other problem you pointed, thanks @LinusU

PR #343, opened 2024-01-23
Optimize hexWrite with an array lookup table

chjj:
Should fix the slowdown introduced by the hash table lookups in #303

My own benchmarks seem to indicate a 10x speedup.

### DIFF
--- a/index.js
+++ b/index.js
@@ -855,19 +855,24 @@ function hexWrite (buf, string, offset, length) {
 
   const strLen = string.length
 
-  if (length > strLen / 2) {
-    length = strLen / 2
+  if (length > (strLen >>> 1)) {
+    length = strLen >>> 1
   }
-  let i
-  for (i = 0; i < length; ++i) {
-    const a = hexCharValueTable[string[i * 2]]
-    const b = hexCharValueTable[string[i * 2 + 1]]
-    if (a === undefined || b === undefined) {
+
+  for (let i = 0; i < length; ++i) {
+    const a = string.charCodeAt(i * 2 + 0)
+    const b = string.charCodeAt(i * 2 + 1)
+    const hi = hexCharValueTable[a & 0x7f]
+    const lo = hexCharValueTable[b & 0x7f]
+
+    if ((a | b | hi | lo) & ~0x7f) {
       return i
     }
-    buf[offset + i] = a << 4 | b
+
+    buf[offset + i] = (hi << 4) | lo
   }
-  return i
+
+  return length
 }
 
 function utf8Write (buf, string, offset, length) {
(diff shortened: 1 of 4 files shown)


chjj:
Wrote a proper benchmark. Here are the numbers:

Without this PR:

```
$ node perf/write-hex.js
BrowserBuffer#write(4096, "hex") x 5,697 ops/sec ±0.29% (91 runs sampled)
NodeBuffer#write(4096, "hex") x 105,578 ops/sec ±0.04% (90 runs sampled)
Fastest is NodeBuffer#write(4096, "hex")
```

With this PR:

```
$ node perf/write-hex.js
BrowserBuffer#write(4096, "hex") x 34,100 ops/sec ±0.07% (93 runs sampled)
NodeBuffer#write(4096, "hex") x 105,699 ops/sec ±0.04% (92 runs sampled)
Fastest is NodeBuffer#write(4096, "hex")
```

Though, on my local machine I was getting an even larger speedup.

index.js
```diff
@@ -855,19 +855,24 @@ function hexWrite (buf, string, offset, length) {
...
   }
-  let i
-  for (i = 0; i < length; ++i) {
-    const a = hexCharValueTable[string[i * 2]]
-    const b = hexCharValueTable[string[i * 2 + 1]]
-    if (a === undefined || b === undefined) {
+
+  for (let i = 0; i < length; ++i) {
+    const a = string.charCodeAt(i * 2 + 0)
+    const b = string.charCodeAt(i * 2 + 1)
```
dcousens, line 864: What happens on unicode?

chjj: Bitwise trickery handles utf16 characters on line 868.

index.js
```diff
@@ -855,19 +855,24 @@ function hexWrite (buf, string, offset, length) {
...
+    const hi = hexCharValueTable[a & 0x7f]
+    const lo = hexCharValueTable[b & 0x7f]
+
+    if ((a | b | hi | lo) & ~0x7f) {
```
chjj, line 868: This line is kind of magic: it's checking for multiple things. If `a` or `b` exceed `0x7f` (ascii), the check will fail. If `a` or `b` result in a lookup of a non-hex character in `hexCharValueTable`, `hi` and/or `lo` become `-1`, also causing the check to fail.

This is the only I know of to handle all possible errors in a single branch. I used this same trick in [bcrypto][1]. It seems to work well.

[1]: https://github.com/bcoin-org/bcrypto/blob/2b9f954/lib/js/base16.js#L94

chjj: Also, `a` and `b` are ANDed by `0x7f` before the array access to prevent the function from being deoptimized due to an OOB access.

chjj: Some examples:

```
> Buffer.from('ab\u2015c', 'hex')
<Buffer ab>
> Buffer.from('ab\u1520c', 'hex')
<Buffer ab>
> Buffer.from('ab\u0715c', 'hex')
<Buffer ab>
```

dcousens: Maybe an easy way to assure this could be to add some tests?

chjj: Tests written. It tests the entire 16 bit space along with another test for odd-length strings.

dcousens: Ha, that's brilliant.  Thanks @chjj

dcousens (approved):
LGTM

chjj:
I added more cases to the test. It's a bit spammy now. Let me know if you want me to reduce that.

dcousens:
Thanks for your work @chjj

PR #345, opened 2024-01-31
Remove class syntax for better babelification (2)

chjj:
Fixes the issues with #344.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1826,42 +1826,36 @@ Buffer.prototype.fill = function fill (val, start, end, encoding) {
 // Simplified versions from Node, changed for Buffer-only usage
 const errors = {}
 function E (sym, getMessage, Base) {
-  errors[sym] = class NodeError extends Base {
-    constructor () {
-      super()
+  function NodeError () {
+    const err = new Base(getMessage.apply(null, arguments))
 
-      Object.defineProperty(this, 'message', {
-        value: getMessage.apply(this, arguments),
-        writable: true,
-        configurable: true
-      })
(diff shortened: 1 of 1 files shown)


dcousens:
I'm not against this, but the specifics of Chrome **7**, which best I can guess is from around 2010? That seems silly?

chjj:
I can give a more realistic reason then: someone who is compiling a bundle and has a babel config set to _anything_ that triggers pre-es2015 shims will receive a class shim from babel. 

The babel class shims do not work when inheriting from any "special" builtin-object (Error, Array, TypedArray are examples). This is because they make the mistake I did originally and transform the constructor to follow the `Base.call(this)` pattern instead of the `Object.setPrototypeOf(new Error(msg), NodeError.prototype)` pattern this PR does.

In short, NodeError will not be an actual Error object if this occurs.

chjj:
In any case, the code as it exists on master is incorrect. `code` should be an enumerable property for example. While this may seem minor, I imagine user's are used to seeing it in the console output:

``` sh
$ node
Welcome to Node.js v21.6.1.
Type ".help" for more information.
> Buffer.alloc(1).write({})
Uncaught TypeError: argument must be a string
    at Buffer.write (node:buffer:1086:17) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

I don't think util.inspect displays the `code` property on errors unless it is enumerable.

dcousens (approved):
LGTM

dcousens:
Thanks for the added context @chjj

Issue #329, opened 2023-07-03
subarray returns Uint8Array which can't handle toString('uff-8')


anion155:
```
import { Buffer } from 'buffer';

declare const messageRaw: ArrayBuffer;
const message = Buffer.from(messageRaw);
const headCorrect = message.slice(0, size).toString('utf-8'); // 57,57,97,102,101,102,56,57,97....
const headIncorrect = message.subarray(0, size).toString('utf-8'); // 99af...
```

personalizedrefrigerator:
Related:
- #260
- https://github.com/feross/buffer/pull/261

chjj:
Any runtime which displays this behavior doesn't implement the es2016 "subarray calls child class" behavior and is returning a plain Uint8Array. 

Node v4.0.0 to v8.0.0 displayed this behavior, even with the native node.js Buffer object. So if you are running with an older version of node.js, it could be argued that feross/buffer is accurately emulating the native node.js Buffer.

This is probably only fixable with an ugly hack. Not sure if people are open to that. Personally, I would just avoid using `subarray` and stick to `slice` when using Buffers.

chjj:
Note, if you are experiencing this in a runtime which *does* support es2016 (i.e. a modern runtime), please post the runtime/browser/platform here. 

If there is actually a popular runtime out there that doesn't implement es2016 properly, we may have to support it.

chjj:
@hedi-edelbloute is this affecting iOS, Android, or both? Which versions?

chjj:
In the meantime you could try this monkey patch:

``` js
if (!(Buffer.alloc(1).subarray(0, 1) instanceof Buffer)) {
  Buffer.prototype.subarray = function subarray() {
    const result = Uint8Array.prototype.subarray.apply(this, arguments);
    Object.setPrototypeOf(result, Buffer.prototype);
    return result;
  };
}
```

hedi-edelbloute:
@chjj My phone is a Redmi note 10 Pro on Android 11, I'm trying to gather more informations. I think it also occured on iOS but i'm going to confirm this.

Thank you for the patch proposal and your help, it's greatly appreciated.

chjj:
Is your Android web view somehow using an old version of chrome? Is there any way to check? I'm not an expert on mobile web views, but I'm certain chrome version 60.x.x included v8 version 6.0.286 which should have the correct es2016 subarray behavior.

hedi-edelbloute:
[...]

PR #346, opened 2024-02-03
Fix issues surrounding type checking for Uint8Arrays and Buffers

chjj:
Logic behind this:

1. `b instanceof Uint8Array` always returns true for Buffers created in the same context.
2. `ArrayBuffer.isView(b)` returns true for both Uint8Arrays and Buffers created in __any__ context (at least in v8 -- hopefully other browsers too).
3. `isInstance(b, Uint8Array)` returns true for Uint8rrays created in __any__ context but currently only returns true for Buffers created in the __same__ context.

Number 3 has an inconsistency. This PR seeks to rectify that and also optimize several functions in doing so.

If we modify the code such that `isInstance(b, Uint8Array)` returns true for Buffers created in __all__ contexts, we can drop a ton of the "buffer casting" for Uint8Arrays (which in most cases is actually entirely redundant and causes a slowdown -- for example, in `Buffer.compare`). I suspect a lot of this code is actually left over from the days when buffer.js still supported IE6.

Not only does this fix a few bugs (e.g. `Buffer#equals` and `Buffer#copy` not accepting Uint8Arrays as their target), but it can also optimize a few functions by avoiding the buffer casts.

In effect, we treat Buffers and Uint8Arrays (from any context) as though they were of the exact same class, which is what node.js does.

### DIFF
--- a/index.js
+++ b/index.js
@@ -312,6 +312,7 @@ function fromArrayBuffer (array, byteOffset, length) {
 
 function fromObject (obj) {
   if (Buffer.isBuffer(obj)) {
+    // Note: Probably not necessary anymore.
     const len = checked(obj.length) | 0
     const buf = createBuffer(len)
 
@@ -358,9 +359,7 @@ Buffer.isBuffer = function isBuffer (b) {
 }
 
 Buffer.compare = function compare (a, b) {
-  if (isInstance(a, Uint8Array)) a = Buffer.from(a, a.offset, a.byteLength)
-  if (isInstance(b, Uint8Array)) b = Buffer.from(b, b.offset, b.byteLength)
-  if (!Buffer.isBuffer(a) || !Buffer.isBuffer(b)) {
+  if (!isInstance(a, Uint8Array) || !isInstance(b, Uint8Array)) {
     throw new TypeError(
       'The "buf1", "buf2" arguments must be one of type Buffer or Uint8Array'
     )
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -358,9 +359,7 @@ Buffer.isBuffer = function isBuffer (b) {
 }
 
 Buffer.compare = function compare (a, b) {
-  if (isInstance(a, Uint8Array)) a = Buffer.from(a, a.offset, a.byteLength)
-  if (isInstance(b, Uint8Array)) b = Buffer.from(b, b.offset, b.byteLength)
```
chjj, line 362: Note that these lines were reallocating the entire buffer __every__ time, regardless of whether the inputs were Uint8Arrays or Buffers. This caused a massive slowdown when comparing large buffers.

index.js
```diff
@@ -644,10 +633,7 @@ if (customInspectSymbol) {
 }
 
 Buffer.prototype.compare = function compare (target, start, end, thisStart, thisEnd) {
-  if (isInstance(target, Uint8Array)) {
-    target = Buffer.from(target, target.offset, target.byteLength)
```
chjj, line 648: Same issue as `Buffer.compare`.

index.js
```diff
@@ -627,7 +617,6 @@ Buffer.prototype.toString = function toString () {
 Buffer.prototype.toLocaleString = Buffer.prototype.toString
 
 Buffer.prototype.equals = function equals (b) {
-  if (!Buffer.isBuffer(b)) throw new TypeError('Argument must be a Buffer')
```
chjj, line 630: This was a bug: Buffers are supposed to be comparable with plain Uint8Arrays.

index.js
```diff
@@ -1714,7 +1697,7 @@ Buffer.prototype.writeDoubleBE = function writeDoubleBE (value, offset, noAssert
 
 // copy(targetBuffer, targetStart=0, sourceStart=0, sourceEnd=buffer.length)
 Buffer.prototype.copy = function copy (target, targetStart, start, end) {
-  if (!Buffer.isBuffer(target)) throw new TypeError('argument should be a Buffer')
```
chjj, line 1717: This was also a bug. Node.js allows copying onto plain Uint8Arrays.

test/typing.js
```diff
@@ -0,0 +1,179 @@
...
+
+// Extracted from the index.js code for testing purposes.
+function isInstance (obj, type) {
+  return (obj instanceof type) ||
+         (obj != null &&
+          obj.constructor != null &&
+          obj.constructor.name != null &&
+          obj.constructor.name === type.name) ||
+         (type === Uint8Array && Buffer.isBuffer(obj))
+}
```
dcousens, line 37: Could probably dedupe, but I digress

index.js
```diff
@@ -312,6 +312,7 @@ function fromArrayBuffer (array, byteOffset, length) {
 
 function fromObject (obj) {
   if (Buffer.isBuffer(obj)) {
+    // Note: Probably not necessary anymore.
```
dcousens, line 315: What isn't necessary about this now?

chjj: Because `ArrayBuffer.isView` returns true for `Uint8Array` subclasses like `Buffer`. So this case will actually be handled earlier, in `Buffer.from`:

``` js
  if (ArrayBuffer.isView(value)) {
    return fromArrayView(value)
  }
```

I left the `Buffer.isBuffer(obj)` check in there because I was concerned that maybe non-v8 javascript engines' `ArrayBuffer.isView` does not return `true` for `Uint8Array` subclasses (from the same context or another). 

Because `ArrayByffer.isView` is so magic in v8, I actually wanted to test it in FF and Safari to make sure they had the same behavior, but I didn't get around to it, so leaving the check in is just me being cautious.

dcousens: You should add comments like ^ into your code :blue_heart:

chjj: Yeah, it's something I need to work on.

hedi-edelbloute:
Hi @chjj, your update looks well ! Thanks

Do you know if it could fix this issue https://github.com/feross/buffer/issues/329 ? :thinking:

chjj:
@hedi-edelbloute, it shouldn't affect `subarray` at all. feross/buffer doesn't ever try to patch or change `subarray`, but if you are experiencing the issue described in #329, please post the affected javascript engine and version in that issue.

hedi-edelbloute:
@chjj I'm going to provide infos on the issue then, thanks

PR #347, opened 2024-02-03
Export missing globals for compat

chjj:


dcousens:
`Buffer` doesn't export these in node? What's the intent?

chjj:
- https://nodejs.org/docs/latest/api/buffer.html#class-blob
- https://nodejs.org/docs/latest/api/buffer.html#class-file
- https://nodejs.org/docs/latest/api/buffer.html#nodebuffer-module-apis

``` sh
$ node
Welcome to Node.js v21.6.1.
Type ".help" for more information.
> require('buffer').Blob
[class Blob]
> require('buffer').File
[class File extends Blob]
> require('buffer').atob
[Function: atob]
> require('buffer').btoa
[Function: btoa]
```

dcousens:
Interesting, I thought this would be equivalent
```bash
node
Welcome to Node.js v21.6.1.
Type ".help" for more information.
> Buffer.atob
undefined
> require('buffer').atob
[Function: atob]
```

Thanks @chjj

dcousens:
Ah, the global scope only has `Buffer`, nevermind

chjj:
This is silly, but I'm now actually having second thoughts on this one. 

I know this is compatible with browserify as browserify exposes `global` to the module, but I'm curious about other bundlers: parcel/webpack/rollup/esbuild/etc (I know the latter two steer their users towards ES modules, but I think they have some kind of CJS compatibility plugins too?).

Maybe we should switch this to a `typeof x !== 'undefined'` pattern to avoid potentially breaking non-browserify bundlers?

Or maybe I'm just being paranoid and they all support `global`.

dcousens:
I'm happy with the cautious approach, but ideally, can we test this package, in our CI, using some other bundlers?

PR #348, opened 2024-02-03
Implement Buffer.copyBytesFrom

chjj:
Implements [Buffer.copyBytesFrom][1].

Also fixes a bug with `ERR_INVALID_ARG_TYPE`.

[1]: https://nodejs.org/docs/latest/api/buffer.html#static-method-buffercopybytesfromview-offset-length

### DIFF
--- a/index.js
+++ b/index.js
@@ -245,6 +245,41 @@ Buffer.allocUnsafeSlow = function (size) {
   return allocUnsafe(size)
 }
 
+/**
+ * Copies the underlying memory of `view` into a new `Buffer`.
+ */
+Buffer.copyBytesFrom = function copyBytesFrom (view, offset, length) {
+  if (!ArrayBuffer.isView(view) || !view.subarray) {
+    throw new errors.ERR_INVALID_ARG_TYPE('view', 'TypedArray', view)
+  }
+
+  if (view.length === 0) return createBuffer(0)
+
+  if (offset !== undefined || length !== undefined) {
+    if (offset !== undefined) {
+      validateInteger(offset, 'offset')
+      if (offset >= view.length) return createBuffer(0)
+    } else {
+      offset = 0
+    }
+
+    let end
+
+    if (length !== undefined) {
+      validateInteger(length, 'length')
+      end = offset + length
+    } else {
+      end = view.length
+    }
+
+    view = view.subarray(offset, end)
+  }
+
+  view = new Uint8Array(view.buffer, view.byteOffset, view.byteLength)
+
+  return fromArrayView(view)
+}
+
 function fromString (string, encoding) {
   if (typeof encoding !== 'string' || encoding === '') {
     encoding = 'utf8'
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -245,6 +245,41 @@ Buffer.allocUnsafeSlow = function (size) {
...
+    } else {
+      end = view.length
+    }
+
+    view = view.subarray(offset, end)
+  }
+
+  view = new Uint8Array(view.buffer, view.byteOffset, view.byteLength)
+
+  return fromArrayView(view)
```
dcousens, line 280:
```suggestion
  if (offset !== undefined || length !== undefined) {
    if (offset !== undefined) validateInteger(offset, 'offset')
    if (length !== undefined) validateInteger(length, 'length')
    const offset_ = offset ?? 0
    const end = (offset + length) ?? view.length
    view = view.subarray(offset, end)
  }
  
  view = new Uint8Array(view.buffer, view.byteOffset, view.byteLength)
  return fromArrayView(view)
```

dcousens (approved):
LGTM, with a nit

test/node/test-buffer-from.js
```diff
@@ -66,3 +67,73 @@ deepStrictEqual(
   throws(() => Buffer.from(input), errMsg);
 });
 
+{
+  const u16 = new Uint16Array([0xffff]);
+  const b16 = Buffer.copyBytesFrom(u16);
+  u16[0] = 0;
+  assert.strictEqual(b16.length, 2);
+  assert.strictEqual(b16[0], 255);
+  assert.strictEqual(b16[1], 255);
```
dcousens, line 76: How are semi's here? :thinking:
Shouldn't the linter break on this?

dcousens: (we can resolve this in another pull request)

index.js
dcousens: Generally, a preference for flat/functional style, even if the actual number of branches is the same

test/node/test-buffer-from.js
chjj: I think it's because I pulled these directly from the node.js tests.

index.js
chjj: This code was ported from the node.js code as faithfully as possible to ensure accurate behavior. I can change it though.

While we're on the subject: what is the minimum ES version we support? Since the `??` operator is an ES2020 feature, it would be good to have clarity on that. BigInt is also an ES2020 feature, but we include graceful degradation for the bigint methods.

test/node/test-buffer-from.js
dcousens: If these are directly from the node tests, can you add an please add a reference?
(and potentially adhere to whatever LICENSE notice may be required there)

chjj: Is that not already accounted for? As far as I can tell, all of the tests in `test/node` are ripped straight from the node.js codebase, right down to the filenames. I even recognize [these ones][1] -- I wrote them. :blush:

[1]: https://github.com/feross/buffer/blob/master/test/node/test-buffer-badhex.js

dcousens: Thanks for hunting that down @chjj, both are MIT, so hopefully this is simply about attribution, I think an inline URL would be helpful too so others can verify the tests against the upstream

PR #352, opened 2024-02-05
Add string decoding benchmarks

chjj:


perf/write-utf8.js
```diff
@@ -0,0 +1,297 @@
...
+var suite2 = util.suite(false)
+var suite3 = util.suite(false)
+
+var LENGTH = 7706
+var browserSubject = BrowserBuffer.alloc(LENGTH)
+var nodeSubject = Buffer.alloc(LENGTH)
+
+// LLM-generated poems in east asian languages.
+// Hopefully a lot of surrogate pairs in here.
+var str = `
```
dcousens, line 15: As someone who doesn't speak either of these languages, and I don't know what the content could be; maybe we could shuffle the string as a random set, and not care about the content?

chjj: I was originally looking at the [Thousand Character Classic][1] -- an ancient chinese text for teaching written chinese, but I decided it wasn't long enough (but we could just `.repeat` it I suppose).

I _did_ plug all of this into google translate. It's some pretty generic and generally positive/optimistic poetry. Nothing weird that I could see.

[1]: https://en.wikipedia.org/wiki/Thousand_Character_Classic

chjj: If you're really worried, I have a friend literate in Japanese and several friends literate in Chinese. I could have them double check it. The Korean part will be harder. I don't know anyone literate in Korean who can read beyond a 2nd grade level.

chjj: I think I'll combine a few things. I'm finding some stuff on wikisource. [The Art of War][1] has a ton of text we can use. I'll find something for other languages too.

[1]: https://zh.wikisource.org/wiki/%E5%AD%AB%E5%AD%90%E5%85%B5%E6%B3%95

PR #355, opened 2024-03-01
Add node.js attribution to LICENSE

chjj:
Should cover any node.js code we decide to re-use (primarily for `test/node` right now).


dcousens (approved):
LGTM

dcousens:
Your question from https://github.com/feross/buffer/pull/348

> Okay, after looking, I guess it isn't accounted for. Should I add a the node.js LICENSE file inside the test/node directory, or should I concatenate the existing LICENSE file with the node.js one?

I don't mind, and as I don't know the history of the project fully, I think covering this broadly is OK, the licenses are the same, with different attribution. I might ping @feross in any event.

PR #363, opened 2024-10-08
README.md - fixed standalone link

canerder:
Previous link: https://bundle.run/buffer <-- webpage displays internal server error
New (working) link: https://bundle.run/buffer@6.0.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ npm install buffer
 This module was previously called **native-buffer-browserify**, but please use **buffer**
 from now on.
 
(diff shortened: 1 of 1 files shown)


dcousens:
lgtm, but internal server error... probably an upstream bug

dcousens:
@rich-harris @xtuc should we still be using this?

Rich-Harris:
To be honest I wasn't aware it was still working at all...@xtuc has worked on it more recently than I have but that's very relative — the last commit was over 4 years ago. You're probably better off using something like https://esm.sh/

PR #366, opened 2024-12-11
Performance gains for hex and base64

jawj:
Following up on #364, I created a new package — [hextreme](https://github.com/jawj/hextreme) — for fast hex and base64 encoding and decoding, and I've updated this library to use it.

The tests are passing, and I've added benchmarks that show performance gains (in Node) of 3 – 13x:

Before:

```
BrowserBuffer#toString("hex") x 72,483 ops/sec ±0.19% (99 runs sampled)
NodeBuffer#toString("hex") x 527,266 ops/sec ±0.33% (97 runs sampled)
Fastest is NodeBuffer#toString("hex")

BrowserBuffer#from(hexString, "hex") x 54,803 ops/sec ±0.84% (95 runs sampled)
NodeBuffer#from(hexString, "hex") x 342,656 ops/sec ±0.14% (97 runs sampled)
Fastest is NodeBuffer#from(hexString, "hex")

BrowserBuffer#toString("base64") x 30,815 ops/sec ±0.18% (94 runs sampled)
NodeBuffer#toString("base64") x 2,594,667 ops/sec ±0.44% (97 runs sampled)
Fastest is NodeBuffer#toString("base64")

BrowserBuffer#from(base64String, "base64") x 28,184 ops/sec ±0.36% (97 runs sampled)
NodeBuffer#from(base64String, "base64") x 1,218,393 ops/sec ±0.40% (96 runs sampled)
Fastest is NodeBuffer#from(base64String, "base64")
```

After:

```
BrowserBuffer#toString("hex") x 348,151 ops/sec ±0.51% (97 runs sampled)
NodeBuffer#toString("hex") x 545,605 ops/sec ±0.60% (96 runs sampled)
Fastest is NodeBuffer#toString("hex")

BrowserBuffer#from(hexString, "hex") x 187,737 ops/sec ±0.29% (95 runs sampled)
NodeBuffer#from(hexString, "hex") x 348,078 ops/sec ±0.24% (98 runs sampled)
Fastest is NodeBuffer#from(hexString, "hex")

BrowserBuffer#toString("base64") x 404,447 ops/sec ±0.27% (98 runs sampled)
NodeBuffer#toString("base64") x 2,658,187 ops/sec ±0.42% (97 runs sampled)
Fastest is NodeBuffer#toString("base64")

BrowserBuffer#from(base64String, "base64") x 92,365 ops/sec ±0.45% (98 runs sampled)
[...]

Issue #364, opened 2024-11-12
Faster `toString('hex')` implementation


jawj:
In a similar vein to #245, I have an implementation of `toString('hex')` that's between 2x and 4x faster than the existing one, depending on the browser/engine.

See: https://jsbench.me/evm3ejel2i/3

Possible drawbacks of this implementation:

* Requires `TextDecoder` (does this package support IE currently?)
* Likely uses more memory, since it allocates a `Uint16Array` to hold intermediate data

Would you accept a PR to switch to this implementation?

jawj:
Thinking about it, we can probably limit the extra memory needed by doing it in chunks ...

jawj:
I've packaged this code up on npm: https://www.npmjs.com/package/hextreme. For multi-megabyte arrays it can be many times faster than the approach currently taken here.

Are you willing to consider a PR to use this package for `.toString('hex')`?

dcousens:
I think reasonably we can track LTS, and with IE officially out of support - I'm happy to say we that we could drop support for that.

jawj:
Thanks. I actually added a fallback implementation where there's no `TextDecoder` anyway, which is still slightly faster than the current approach because it does a bit of loop unrolling.

I also limited the biggest possible temporary buffer allocation to 1MB, by doing the conversion in chunks, which makes very little difference in speed.

I'll go ahead and submit a PR in the next few days then.

PR #367, opened 2024-12-12
Update `hexSlice` to use TypedArrays and TextDecoder for large speed gains

jawj:
Following up on #364 and #366, this patch switches the `hexSlice` implementation to use TypedArrays and TextDecoder. It also adds a performance benchmark for `toString('hex')`.

In Node on an M3 Pro MacBook, the benchmark shows a 5x speed-up on a 4096-byte buffer, achieving 75% of the speed of the native method. The speed-up is generally larger as the buffer size increases.

Before:

```
BrowserBuffer#toString("hex") x 72,483 ops/sec ±0.19% (99 runs sampled)
NodeBuffer#toString("hex") x 527,266 ops/sec ±0.33% (97 runs sampled)
Fastest is NodeBuffer#toString("hex")
```

After:

```
BrowserBuffer#toString("hex") x 403,058 ops/sec ±0.30% (98 runs sampled)
NodeBuffer#toString("hex") x 533,325 ops/sec ±0.21% (98 runs sampled)
Fastest is NodeBuffer#toString("hex")
```


jawj:
@dcousens Any chance of merging this in the near future? Thanks.

PR #373, opened 2025-09-20
Replace BigInt function types with the correct bigint primitive types

ItsFrankieD:
The types for bigint-related methods took and returned `BigInt` functions rather than the intended `bigint` primitive types. I figured these methods are used quite sparsely, so this never got caught!


dcousens:
Thanks @ItsFrankieD 💛

